PR #31, opened 2020-05-20
adds wizard-remove-props rule

### LINKED ISSUE #28: Add wizard-remove-props
## wizard-remove-props [(#4142)](https://github.com/patternfly/patternfly-react/pull/4142)
- [ ] **WizardNav**: Removed prop `isCompactNav `. This prop is no longer used.
- [ ] **Wizard:** Removed inPage prop. By default the Wizard will be displayed in page, filling its parent container. If the consumer passes a managed `isOpen` flag, then the Wizard will be displayed in a modal.
- [ ] **Wizard:** Remove isFullHeight and isFullWidth props, all instances of them should be removed from your application. [(#4116)](https://github.com/patternfly/patternfly-react/pull/4116)

jenny-s51:
Closes #28

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/rules/wizard-remove-props.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/wizard-remove-props.js
@@ -1,0 +1,35 @@
+const { getPackageImports } = require('../helpers');
+
+const renames = {
+  'isCompactNav': 'REMOVE',
+  'inPage': 'REMOVE',
+  'isFullWidth': 'REMOVE',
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin-pf-codemods/lib/rules/wizard-remove-props.js
```diff
@@ -0,0 +1,35 @@
+const { getPackageImports } = require('../helpers');
+
+const renames = {
```
redallen, line 3: This could just be a list, but since `Boolean('REMOVE') == true`, you're okay :)

PR #57, opened 2020-05-28
Nav list variant

### LINKED ISSUE #12: Add nav-list-variant
## nav-list-variant [(#4225)](https://github.com/patternfly/patternfly-react/pull/4225)
- [ ] **Nav:** `variant` prop has been removed from `NavList`. Pass variant={`horizontal` | `tertiary`} to `Nav` instead.

ddonahue007:
Closes #12 

Fixes for NavList variant property being moved to Nav component.

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/rules/nav-list-variant.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/nav-list-variant.js
@@ -8,32 +8,51 @@ module.exports = {
     const navImport = imports.find(imp => imp.imported.name === 'Nav');
     const NavListImport = imports.find(imp => imp.imported.name === 'NavList');
 
-    return !navImport || !NavListImport ? {} : {
+    return !NavListImport ? {} : {
       JSXElement(node) {
+
         if (NavListImport.local.name === node.openingElement.name.name) {
-          const hasNavParent = node.parent
-              && node.parent.openingElement.name.name === navImport.local.name
-              && node.parent.children.filter(child => child.type === 'JSXElement').length === 1;
+          let hasNavParent;
+          let navImportName = "Nav";
+
+          if (navImport) {
+            hasNavParent = node.parent
+                && node.parent.openingElement.name.name === navImport.local.name
+                && node.parent.children.filter(child => child.type === 'JSXElement').length === 1;
+
+            navImportName = navImport.local.name;
+          }
+
           const variantAttr = node.openingElement.attributes.find(attribute => {
-              return attribute.name.name === 'variant'
+            return attribute.name.name === 'variant'
           });
 
-          if (variantAttr && variantAttr.value !== null) {
-            const variantVal = context.getSourceCode().getText(variantAttr);
-            context.report({
-              node,
-              message: `variant has been removed from ${node.openingElement.name.name}, use <${navImport.local.name} ${variantVal}> instead`,
-              fix(fixer) {
-                const fixes = [fixer.replaceText(variantAttr, '')];
-                if(hasNavParent) {
-                  fixes.push(fixer.insertTextAfter(node.parent.openingElement.name, ' ' + variantVal))
+          const variantVal = context.getSourceCode().getText(variantAttr) || '"horizontal" | "default" | "tertiary"';
+
+          console.log("Variant: " + variantVal);
+
+          if (variantAttr) {
+            if (variantAttr.value !== null) {
+              context.report({
+                node,
+                message: `variant has been removed from ${node.openingElement.name.name}, use <${navImportName} ${variantVal}> instead`,
+                fix(fixer) {
+                  const fixes = [fixer.replaceText(variantAttr, '')];
+                  if (hasNavParent) {
+                    fixes.push(fixer.insertTextAfter(node.parent.openingElement.name, ' ' + variantVal))
+                  }
+                  return fixes;
                 }
-                return fixes;
-              }
-            });
+              });
+            } else {
+              context.report({
+                node,
+                message: `variant has been removed from ${node.openingElement.name.name}, use <${navImportName} variant={"horizontal" | "default" | "tertiary"}> instead`,
+              });
+            }
           }
         }
       }
-    };
+    }
   }
-};
+}
--- a/packages/eslint-plugin-pf-codemods/lib/rules/nav-list-variant.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/nav-list-variant.js
@@ -18,16 +18,15 @@ module.exports = {
               return attribute.name.name === 'variant'
           });
 
-          if (variantAttr) {
-            const variantVal = '"horizontal" || "default" || "tertiary"';
-
+          if (variantAttr && variantAttr.value !== null) {
+            const variantVal = context.getSourceCode().getText(variantAttr) || '"horizontal" | "default" | "tertiary"';
             context.report({
               node,
-              message: `variant has been removed from ${node.openingElement.name.name}, use <${navImport.local.name} variant=${variantVal}> instead`,
+              message: `variant has been removed from ${node.openingElement.name.name}, use <${navImport.local.name} ${variantVal}> instead`,
               fix(fixer) {
                 const fixes = [fixer.replaceText(variantAttr, '')];
-                if(hasNavParent ){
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin-pf-codemods/lib/rules/nav-list-variant.js
```diff
@@ -0,0 +1,40 @@
...
+        if (NavListImport.local.name === node.openingElement.name.name) {
+          const hasNavParent = node.parent
+              && node.parent.openingElement.name.name === navImport.local.name
+              && node.parent.children.filter(child => child.type === 'JSXElement').length === 1;
+          const variantAttr = node.openingElement.attributes.find(attribute => {
+              return attribute.name.name === 'variant'
+          });
+
+          if (variantAttr) {
+            const variantVal = '"horizontal" || "default" || "tertiary"';
```
redallen, line 22: Can you change this to `const variantVal = context.getSourceCode().getText(variantAttr) || '"horizontal" | "default" | "tertiary"';`?

packages/eslint-plugin-pf-codemods/lib/rules/nav-list-variant.js
```diff
@@ -0,0 +1,40 @@
...
+
+            context.report({
+              node,
+              message: `variant has been removed from ${node.openingElement.name.name}, use <${navImport.local.name} variant=${variantVal}> instead`,
+              fix(fixer) {
+                const fixes = [fixer.replaceText(variantAttr, '')];
+                if(hasNavParent ){
+                  fixes.push(fixer.insertTextAfter(node.parent.openingElement.name, ' ' + context.getSourceCode().getText(variantAttr)))
```
redallen, line 30: Change to `' ' + variantVal`

packages/eslint-plugin-pf-codemods/test/rules/nav-list-variant.js
```diff
@@ -0,0 +1,75 @@
...
+<Nav variant="horizontal">
+    <NavList >
+        some item
+    </NavList>
+</Nav>`,
+      errors: [{
+        message: `variant has been removed from NavList, use <Nav variant=${variantVal}> instead`,
+        type: "JSXElement",
+      }]
+    }
```
redallen, line 71: Can you add this silly test case?
```jsx
import { Nav, NavList } from '@patternfly/react-core';
<Nav>
    <NavList variant>
        some item
    </NavList>
</Nav>
```

I think currently this rule would not produce valid JSX since `context.getSourceCode().getText(variantAttr)` would return an empty string.

packages/eslint-plugin-pf-codemods/lib/rules/nav-list-variant.js
ddonahue007: done

packages/eslint-plugin-pf-codemods/lib/rules/nav-list-variant.js
ddonahue007: I removed the '|| '"horizontal" | "default" | "tertiary"' as none of the test cases are now using that value.

packages/eslint-plugin-pf-codemods/lib/rules/nav-list-variant.js
```diff
@@ -0,0 +1,39 @@
...
+
+// https://github.com/patternfly/patternfly-react/pull/4225
+module.exports = {
+  create: function(context) {
+    const imports = getPackageImports(context, '@patternfly/react-core')
+        .filter(specifier => ['NavList', 'Nav'].includes(specifier.imported.name));
+    const navImport = imports.find(imp => imp.imported.name === 'Nav');
+    const NavListImport = imports.find(imp => imp.imported.name === 'NavList');
+
+    return !navImport || !NavListImport ? {} : {
```
jschuler, line 11: What if NavList is used in a separate file without the Nav import? Then the rule won't run right

packages/eslint-plugin-pf-codemods/lib/rules/nav-list-variant.js
```diff
@@ -0,0 +1,39 @@
...
+      JSXElement(node) {
+        if (NavListImport.local.name === node.openingElement.name.name) {
+          const hasNavParent = node.parent
+              && node.parent.openingElement.name.name === navImport.local.name
```
jschuler, line 15: will probably error in this case because it is not wrapped in a parent node
File 1: 
```
import { Nav } from '@patternfly/react-core';
import { MyListComponent } from './myList';

<Nav><MyListComponent /></Nav>
```
myList.tsx:
```
import { NavList } from '@patternfly/react-core';

<NavList variant="horizontal" />
```

packages/eslint-plugin-pf-codemods/lib/rules/nav-list-variant.js
ddonahue007: That is correct.

packages/eslint-plugin-pf-codemods/lib/rules/nav-list-variant.js
ddonahue007: Done

packages/eslint-plugin-pf-codemods/lib/rules/nav-list-variant.js
```diff
@@ -0,0 +1,62 @@
...
+                  }
+                  return fixes;
+                }
+              });
+            } else {
+              context.report({
+                node,
+                message: `variant has been removed from ${node.openingElement.name.name}, use <${navImportName} variant={"horizontal" | "default" | "tertiary"}> instead`,
+                fix(fixer) {
+                  const fixes = [fixer.replaceText(variantAttr, '')];
```
jschuler, line 45: i think this might be dangerous since it would simply remove the attribute, how about instead we comment it out instead and even add some helpful text in the comment, such as
```
fixer.replaceText(variantAttr, `/*TODO: move to Nav - ${variantVal}*/`)
```

packages/eslint-plugin-pf-codemods/lib/rules/nav-list-variant.js
```diff
@@ -0,0 +1,62 @@
...
+      JSXElement(node) {
+
+        if (NavListImport.local.name === node.openingElement.name.name) {
+          let hasNavParent;
+          let navImportName = "Nav";
+
+          if (navImport) {
+            hasNavParent = node.parent
+                && node.parent.openingElement.name.name === navImport.local.name
+                && node.parent.children.filter(child => child.type === 'JSXElement').length === 1;
```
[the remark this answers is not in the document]
redallen: To make sure it's the only `<NavList>` child. If there's
```
<Nav>
  <NavList variant="1" />
  <NavList variant="2" />
</Nav>
```
we don't want to add `<Nav variant="123">` at all

packages/eslint-plugin-pf-codemods/test/rules/nav-list-variant.js
```diff
@@ -0,0 +1,125 @@
...
+      errors: [{
+        message: `variant has been removed from NavList, use <Nav variant={"horizontal" | "default" | "tertiary"}> instead`,
+        type: "JSXElement",
+      }]
+    },
+    {
+      code: `import { NavList } from '@patternfly/react-core';
+<NavList variant="horizontal">
+    some item
+</NavList>`,
```
redallen, line 112: Thanks for adding this

packages/eslint-plugin-pf-codemods/test/rules/nav-list-variant.js
```diff
@@ -0,0 +1,125 @@
...
+    <NavList variant>
+        some item
+    </NavList>
+</Nav>`,
+      output: `import { Nav, NavList } from '@patternfly/react-core';
+<Nav>
+    <NavList >
+        some item
+    </NavList>
+</Nav>`,
```
redallen, line 88: Awesome!

packages/eslint-plugin-pf-codemods/lib/rules/nav-list-variant.js
redallen: If we do this, we should also do it `dropdown-rename-icon.js`. I'm okay with opening a separate issue and discussing what's best there.

PR #68, opened 2020-05-29
Nav list variant

ddonahue007:
changes for when there is a disjointed NavList 
Updated rule to use metadata


packages/eslint-plugin-pf-codemods/lib/rules/nav-list-variant.js
```diff
@@ -28,21 +35,35 @@ module.exports = {
...
+                  variantVal: variantVal
+                },
                 fix(fixer) {
-                  const fixes = [fixer.replaceText(variantAttr, '')];
+                  const fixes = [];
                   if (hasNavParent) {
+                    fixes.push(fixer.replaceText(variantAttr, ''));
                     fixes.push(fixer.insertTextAfter(node.parent.openingElement.name, ' ' + variantVal))
+                  } else {
+                    fixes.push(fixer.replaceText(variantAttr, `/*TODO: move to Nav - ${variantVal}*/`));
```
jschuler, line 50: before possible values for variant included: `'default' | 'simple' | 'horizontal' | 'tertiary'`
If the user had `default` or `simple` before we can just remove it, otherwise comment it out

packages/eslint-plugin-pf-codemods/lib/rules/nav-list-variant.js
```diff
@@ -28,21 +35,35 @@ module.exports = {
...
+                messageId: "removeNavListVariant",
+                data: {
+                  navListName: node.openingElement.name.name,
+                  navName: navImportName,
+                  variantVal: variantVal
+                },
                 fix(fixer) {
-                  const fixes = [fixer.replaceText(variantAttr, '')];
+                  const fixes = [];
                   if (hasNavParent) {
```
jschuler, line 46: need to add checks if user used `simple` variant before, since that doesn't map over to the new Nav variant prop

ddonahue007: Done

redallen: Did you push @ddonahue007 ?

packages/eslint-plugin-pf-codemods/lib/rules/nav-list-variant.js
ddonahue007: Done

packages/eslint-plugin-pf-codemods/lib/rules/nav-list-variant.js
ddonahue007: I thought i pushed, I just did another push see what happens

PR #69, opened 2020-05-29
Chore/64 expandable rename expandablesection

### LINKED ISSUE #64: Add expandable-rename-expandablesection
The component Expandable has been renamed to ExpandableSection.

evwilkin:
Closes #64 

Renames import and usage of `Expandable` to `ExpandableSEction`.

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/rules/expandable-rename-expandablesection.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/expandable-rename-expandablesection.js
@@ -1,51 +1,9 @@
-const { getPackageImports } = require('../helpers');
+const { renameComponent } = require('../helpers');
 
 // https://github.com/patternfly/patternfly-react/pull/4116
 module.exports = {
-  create: function(context) {
-    const imports = getPackageImports(context, '@patternfly/react-core')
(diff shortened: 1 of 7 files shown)


packages/eslint-plugin-pf-codemods/lib/rules/expandable-rename-expandablesection.js
```diff
@@ -0,0 +1,51 @@
...
+            fix(fixer) {
+              const fixes = [
+                fixer.replaceText(openingElement, updateTagName(openingElement)),
+                fixer.replaceText(closingElement, updateTagName(closingElement))
+              ];
+              return fixes;
+            }
+          });
+        }
+      }
```
redallen, line 46: Can you make this the default implementation of the `renameComponent` helper and then use it?

https://github.com/patternfly/pf-codemods/blob/master/packages/eslint-plugin-pf-codemods/lib/helpers.js#L83

evwilkin: Thanks for the suggestion - updated this!

evwilkin:
> JK, you have some snapshots to update and a `chipButton` test case to fix

Caught this right after resolving your previous feedback, should be good now - sorry for the quick acting before checks completed!

PR #71, opened 2020-06-02
page header tools rules

### LINKED ISSUE #27: Add use-page-header-tools
## use-page-header-tools [(#4223)](https://github.com/patternfly/patternfly-react/pull/4223)
- [ ] **PageHeader**: `avatar` prop was removed. Add the avatar to the `PageHeaderTools` component instead (which is passed into `PageHeader` via the `headerTools` prop.
- [ ] **PageHeader**: `toolbar` prop was removed. Use the `headerTools` prop instead. Also, if you previously used the `Toolbar`, `ToolbarGroup`, or `ToolbarItem` components for the header, replace them with the `PageHeaderTools`, `PageHeaderToolsGroup`, and `PageHeaderToolsItem` components.

jschuler:
Closes #27


packages/eslint-plugin-pf-codemods/lib/rules/use-page-header-tools.js
```diff
@@ -0,0 +1,71 @@
+const { getPackageImports } = require('../helpers');
```
redallen, line 1: Can we rename this rule to `pageheader-move-avatar`?

PR #73, opened 2020-06-03
now able to specify rules on cli

ddonahue007:
Example of one rule: (remove-unused-imports)
Screen Shot 2020-06-03 at 11 34 24 AM

Example of two rules: (remove-unused-imports background-image-src-enum)
Screen Shot 2020-06-03 at 11 37 39 AM

No rules:
Screen Shot 2020-06-03 at 11 38 47 AM

### DIFF
--- a/packages/pf-codemods/index.js
+++ b/packages/pf-codemods/index.js
@@ -63,7 +63,14 @@ function main() {
     const cliRules = process.argv.splice(process.argv.indexOf('--rules') + 1, 1)[0].split(' ');
 
     for (let i = 0; i < cliRules.length; i++) {
-      finalRules[cliRules[i]]= require('../eslint-plugin-pf-codemods/lib/rules/' + cliRules[i]);
+      try {
+        finalRules[cliRules[i]] = require('../eslint-plugin-pf-codemods/lib/rules/' + cliRules[i]);
+
+      } catch (error) {
+        const iRule = error.message.split("/");
+        console.error("Invalid Rule: " + iRule[iRule.lastIndexOf('rules') + 1])
+        return 2;
+      }
     }
 
     currentOptions.env = { browser: true, node: true,es6: true };
--- a/packages/pf-codemods/index.js
+++ b/packages/pf-codemods/index.js
@@ -2,7 +2,6 @@
 const path = require('path');
 const options = require('eslint/lib/options');
 const { CLIEngine } = require('eslint/lib/cli-engine');
-const lodash = require('lodash');
 
 /**
  * Outputs the results of the linting.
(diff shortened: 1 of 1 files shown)


packages/eslint-plugin-pf-codemods/index.js
```diff
@@ -1,47 +1,59 @@
...
-  "form-fix-isValid": require('./lib/rules/form-fix-isValid'),
-  "expandable-rename-expandablesection": require('./lib/rules/expandable-rename-expandablesection'),
-  "rename-toolbar-components": require('./lib/rules/rename-toolbar-components'),
-  "page-header-prop-rename": require('./lib/rules/page-header-prop-rename'),
-  "page-header-move-avatar": require('./lib/rules/page-header-move-avatar'),
-};
+let rules = {};
+let arguments = process.argv;
+
+if (arguments.includes('--rules')) {
```
redallen, line 4: Can we include this login in the pf-codemods package instead? I don't think the eslint plugin should care what CLI options were passed.

packages/pf-codemods/index.js
```diff
@@ -43,8 +44,15 @@ function printResults(engine, results, format) {
 
 function main() {
   let currentOptions;
+
   try {
-    currentOptions = options.parse(process.argv);
+    let arguments = lodash.cloneDeep(process.argv);
```
redallen, line 49: I think `[...process.argv]` works? No need for lodash

packages/pf-codemods/index.js
```diff
@@ -43,8 +44,15 @@ function printResults(engine, results, format) {
...
+
+    // remove --rules for parsing purposes
+    if (process.argv.includes('--rules')) {
+      arguments.splice(arguments.indexOf('--rules'),2);
+    }
+    currentOptions = options.parse(arguments);
   } catch (error) {
     console.error(error.message);
     return 2;
```
redallen, line 58: Line 62-83 is what needs editing. `rules: undefined` needs to be populated and the `.eslintrc.json` might need to be modified to not use the recommended rules from the plugin (aka ALL rules).

packages/pf-codemods/index.js
```diff
@@ -2,6 +2,7 @@
 const path = require('path');
 const options = require('eslint/lib/options');
 const { CLIEngine } = require('eslint/lib/cli-engine');
+const lodash = require('lodash');
```
dlabrecq, line 5: It's possible `lodash` was installed via another package (e.g., Victory or react-charts), but should add `lodash` as a `pf-codemods` dependency.

packages/pf-codemods/index.js
ddonahue007: I was trying to be safe with using lodash.deepcopy(), but I am sure it is over kill for this. I have changed back to the above suggestion.

packages/pf-codemods/index.js
ddonahue007: I removed the need to use lodash as it is most likely over kill for our needs here.

packages/pf-codemods/index.js
ddonahue007: Done

Issue #78, opened 2020-06-03
breakpointMods not getting fixed


jschuler:
```
<Flex breakpointMods={[{ modifier: 'justify-content-space-between' }] as any}>
  <FlexItem>
    <ResourceLink
      inline
      kind={referenceForModel(PipelineModel)}
      name={name}
      namespace={namespace}
    />
  </FlexItem>
  <FlexItem>
    <TriggerLastRunButton pipelineRuns={pipelineRuns} namespace={namespace} />
  </FlexItem>
</Flex>
```

redallen:
Closed in #87

PR #87, opened 2020-06-04
Fix breakpointMods don't apply

evwilkin:
Towards #78 

This PR updated the breakpointMods codemod in the case of typing the breakpointMods array:
`<Flex breakpointMods={[{ modifier: 'justify-content-space-between' }] as any}></Flex>`


redallen:
Checking out your branch now to fix mocha

redallen:
Pushed a fix to add the parser to the rule tester

PR #91, opened 2020-06-04
Add data-codemods attribute on rename

### LINKED ISSUE #82: Running codemods a second time introduces some issues
Either we need to specify that codemods should only be run once, or do a better job of checking if it's already been run. For example, first time it' run it might correctly replace DataToolbar with Toolbar, and Toolbar with PageHeaderTools, but on second run it would wrongly replace Toolbar again with PageHeaderTools

evwilkin:
Closes #82 

This PR adds data-codemods attributes to renamed components that would potentially be renamed a second time, and leaves removing the previous names' imports to `no-unused-imports` rule (#90)

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/helpers.js
+++ b/packages/eslint-plugin-pf-codemods/lib/helpers.js
@@ -88,26 +88,9 @@ function renameComponent(
     const imports = getPackageImports(context, '@patternfly/react-core')
       .filter(specifier => Object.keys(componentMap).includes(specifier.imported.name));
     const importedNamesArr = imports.map(imp => imp.imported.name);
+    const hasDataCodemodsAttr = [];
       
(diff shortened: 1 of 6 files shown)


packages/eslint-plugin-pf-codemods/lib/helpers.js
```diff
@@ -82,52 +82,48 @@ function renameProp(components, propMap, message, replaceAttribute) {
...
-            node,
-            message: message(importedName, newName),
-            fix(fixer) {
-              return fixer.replaceText(node, newName);
-            }
-          });
-        }
+      ImportDeclaration(node) {
+        ensureImports(context, node, package, Object.values(componentMap));
       },
```
redallen, line 96: You have to add `ensureImports` unconditionally in `ImportDeclaration(node)`

PR #92, opened 2020-06-05
fix for specifying rule(s) on the cli

ddonahue007:
when providing rules the rules should run correctly. This fix allows this to happen.

Single Rule:
Screen Shot 2020-06-05 at 10 09 51 AM

Multiple Rule:
Screen Shot 2020-06-05 at 10 11 19 AM

### DIFF
--- a/packages/pf-codemods/index.js
+++ b/packages/pf-codemods/index.js
@@ -58,9 +58,10 @@ function main() {
   }
 
   // construct rules from cli, if provide
-  let finalRules = [];
+  let finalRules = {};
   if (process.argv.includes('--rules')) {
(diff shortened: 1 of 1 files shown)


packages/pf-codemods/index.js
```diff
@@ -73,10 +74,11 @@ function main() {
       }
     }
 
+    currentOptions.configFile = undefined;
```
redallen, line 77: Can we just do away with the config file entirely on lines 87-89? There's not that much config actually in it...

ddonahue007: Yes we can, I can change it here, but I have it in the --exclude switch changes I am making.

PR #173, opened 2022-12-13
Repo structure and Toolbar codemod

### LINKED ISSUE #155: Toolbar - remove deprecated visiblity prop
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/8212

The `visiblity` prop was deprecated and should be removed from consumers' `Toolbar` code and replaced with the correctly spelled `visibility` prop.

_Required actions:_
1. Build codemod
2. Build test

kmcfaul:
Closes #155 and takes a crack at reorganizing the repo a little (old rules and tests are moved into a `v4` folder, the new rule/test is under a `v5` folder). Also fixes the v4 tests (missing a metadata object) and updates the test commands: `test` will run all tests, `test:v4` will run only the v4 tests, and `test:v5` will run only the v5 tests. 

LMK if we end up wanting another repo layout and I can update or back out the changes save for the new codemod.


nicolethoen (commented):
Lovely! Not sure if we want to restructure or rename the current README file so that it reflects these changes and new test commands?
We might want to rename it to v4-README and then start a clean README for v5 so we can use it to document all changes as we did before?

thatblindgeye (approved):
This is great! Tests pass running locally 💪🏼 No other comments other than I'd be in agreement with Nicole about maybe renaming the current README and adding a fresh one for v5. Depending on if we go that route I can reapprove then

gitdallas (approved):

gitdallas (approved):

thatblindgeye (approved):

nicolethoen (approved):

packages/eslint-plugin-pf-codemods/test/rules/v5/toolbar-remove-visiblity.js
```diff
@@ -0,0 +1,24 @@
...
+const rule = require('../../../lib/rules/v5/toolbar-remove-visiblity');
+
+ruleTester.run("toolbar-remove-visiblity", rule, {
+  valid: [
+    {
+      code: `import { ToolbarContent } from '@patternfly/react-core'; <ToolbarContent visibility />`,
+    },
+    {
+      // No @patternfly/react-core import
+      code: `<ToolbarContent visibility />`,
```
gitdallas, line 11: I noticed this after... but the nonpf test is actually to test the changes don't affect nonpf components of the same name. I've added the small change to my existing pr https://github.com/patternfly/pf-codemods/pull/174/files#diff-287295441f7c3064650362745a50c5f18fa112fc92dcd0c7e3aca3450ccc6f4dR11

PR #175, opened 2022-12-16
feat(FileUpload): remove onChange prop

### LINKED ISSUE #135: FileUpload: removed deprecated onChange prop
Original issue: https://github.com/patternfly/patternfly-react/pull/8155

Description:  We've removed the `onChange` prop completely with no replacement, should remove it for consumers and alert them to this change.

_Required actions:_
1. Build codemod
2. Build test

thatblindgeye:
Closes #135

### DIFF
--- a/packages/eslint-plugin-pf-codemods/test/rules/v5/fileUpload-remove-onChange.js
+++ b/packages/eslint-plugin-pf-codemods/test/rules/v5/fileUpload-remove-onChange.js
@@ -7,7 +7,7 @@ ruleTester.run("fileUpload-remove-onChange", rule, {
       code: `import { FileUpload } from '@patternfly/react-core'; <FileUpload />`,
     },
     // No @patternfly/react-core import
-    { code: "<FileUpload />" },
+    { code: "<FileUpload onChange />" },
   ],
   invalid: [
     {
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin-pf-codemods/test/rules/v5/fileUpload-remove-onChange.js
```diff
@@ -0,0 +1,24 @@
+const ruleTester = require("../../ruletester");
+const rule = require("../../../lib/rules/v5/fileUpload-remove-onChange");
+
+ruleTester.run("fileUpload-remove-onChange", rule, {
+  valid: [
+    {
+      code: `import { FileUpload } from '@patternfly/react-core'; <FileUpload />`,
+    },
+    // No @patternfly/react-core import
+    { code: "<FileUpload />" },
```
gitdallas, line 10: This should include the attribute, the reason for these tests is that we won't touch it if a user has a FileUpload that is not from PF.

gitdallas (changes requested):
small change

packages/eslint-plugin-pf-codemods/test/rules/v5/fileUpload-remove-onChange.js
gitdallas: @kmcfaul  - I just noticed the same thing for your visiblity test, although the chances of another `ToolbarContent` with a wrongly spelled "visiblity" attribute is... pretty small

thatblindgeye: Thanks for catching this!

gitdallas (approved):

PR #178, opened 2022-12-16
feat(misc) - remove removeFindDomNode prop

### LINKED ISSUE #161: Various - removeFindDomNode prop was removed
Follow up to breaking change PR #8316 as well as #8371

`removeFindDomNode` was removed and has been made the default implementation. This might be more of an epic but the codemods I would think would be very similar to each other if it couldn't be done in a single codemod.

Affected components which need a codemod to remove the prop:
- [ ] ApplicationLauncher
- [ ] ClipboardCopy
- [ ] ContextSelector
- [ ] Dropdown
- [ ] NavItem
- [ ] OptionsMenu
- [ ] Popover
- [ ] SearchInput
- [ ] Select
- [ ] OverflowTab
- [ ] Timepicker
- [ ] Tooltip
- [ ] Truncate

_Required actions:_
1. Build codemod
2. Build test

kmcfaul:
Closes #161

### DIFF
--- a/packages/eslint-plugin-pf-codemods/test/rules/v5/remove-removeFindDomNode.js
+++ b/packages/eslint-plugin-pf-codemods/test/rules/v5/remove-removeFindDomNode.js
@@ -1,0 +1,204 @@
+const ruleTester = require('../../ruletester');
+const rule = require('../../../lib/rules/v5/remove-removeFindDomNode');
+
+ruleTester.run("remove-removeFindDomNode", rule, {
+  valid: [
+    {
+      code: `import { ApplicationLauncher } from '@patternfly/react-core'; <ApplicationLauncher />`,
+    },
+    {
+      // No @patternfly/react-core import
+      code: `<ApplicationLauncher />`,
+    },
+    {
+      code: `import { ClipboardCopy } from '@patternfly/react-core'; <ClipboardCopy />`,
+    },
+    {
+      // No @patternfly/react-core import
+      code: `<ClipboardCopy />`,
+    },
+    {
+      code: `import { ContextSelector } from '@patternfly/react-core'; <ContextSelector />`,
+    },
+    {
+      // No @patternfly/react-core import
+      code: `<ContextSelector />`,
+    },
+    {
+      code: `import { Dropdown } from '@patternfly/react-core'; <Dropdown />`,
+    },
+    {
+      // No @patternfly/react-core import
+      code: `<Dropdown />`,
+    },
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin-pf-codemods/test/rules/v5/remove-removeFindDomNode.js
```diff
@@ -0,0 +1,204 @@
...
+const rule = require('../../../lib/rules/v5/remove-removeFindDomNode');
+
+ruleTester.run("remove-removeFindDomNode", rule, {
+  valid: [
+    {
+      code: `import { ApplicationLauncher } from '@patternfly/react-core'; <ApplicationLauncher />`,
+    },
+    {
+      // No @patternfly/react-core import
+      code: `<ApplicationLauncher />`,
```
gitdallas, line 11: The "no pf import" tests actually are to test that non-pf components are not impacted, so the property should be in there.

gitdallas (changes requested):

packages/eslint-plugin-pf-codemods/test/rules/v5/remove-removeFindDomNode.js
gitdallas: Also, should the readme be updated?

kmcfaul: Yep, I figured that distinction out a bit later as I was testing another codemod haha. I'll update the test and also the readme.

gitdallas (approved):
i'd have mapped out the renames and rules just so it's easier to see what's all there and reduce chance of missing something, but otherwise lgtm

packages/pf-codemods/README.md
```diff
@@ -40,6 +40,48 @@ Options:
 
 These rules are based off the breaking change notes for React. Each rule links the breaking change patternfly-react PR in case you want to better understand the change. Also, each rule makes sure you're using a PatternFly component before running.
 
+### remove-removeFindDomNode [(#8371)](https://github.com/patternfly/patternfly-react/pull/8371) [(#8316)](https://github.com/patternfly/patternfly-react/pull/8316)
```
thatblindgeye, line 43: Could be worth placing alphabetically after resolving the conflicts

thatblindgeye (commented):
Looks good! Had a comment below, but otherwise just needs conflicts resolved

PR #179, opened 2022-12-16
feat(ResizeObserver): added new parameter

### LINKED ISSUE #137: ResizeObserver - add new function param
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/8324

The `getResizeObserver` function has a new optional third parameter called `useRequestAnimationFrame`, which likely should be  added and defaulted to true for any consumers using the `getResizeObserver` function. 

_Required actions:_
1. Build codemod
2. Build test

thatblindgeye:
Closes #137

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/resizeObserver-function-param.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/resizeObserver-function-param.js
@@ -45,15 +45,6 @@ module.exports = {
                     },
                   },
                 ],
-                // fix(fixer) {
-                //   if (node.arguments.length !== 3) {
-                //     const lastToken = context.getLastToken(node);
-                //     const fixes = [
-                //       fixer.insertTextBefore(lastToken, ", false"),
-                //     ];
-                //     return fixes;
-                //   }
-                // },
               });
             }
           },
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/resizeObserver-function-param.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/resizeObserver-function-param.js
@@ -1,0 +1,62 @@
+const { getPackageImports } = require("../../helpers");
+
+// https://github.com/patternfly/patternfly-react/pull/8324
+module.exports = {
+  meta: { hasSuggestions: true },
+  create: function (context) {
+    const imports = getPackageImports(context, "@patternfly/react-core").filter(
+      (specifier) => specifier.imported.name === "getResizeObserver"
+    );
+
+    return imports.length == 0
+      ? {}
+      : {
+          CallExpression(node) {
+            if (
+              node.callee.name === "getResizeObserver" &&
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin-pf-codemods/lib/rules/v5/resizeObserver-function-param.js
```diff
@@ -0,0 +1,53 @@
...
+              node.arguments.length !== 3
+            ) {
+              context.report({
+                node,
+                message: `A third parameter, useRequestAnimationFrame, has been added to the getResizeObserver function.`,
+                suggest: [
+                  {
+                    desc: "Pass in a value of false when the callback passed in is debounced. This maintains the previous functionality.",
+                    fix: function (fixer) {
+                      if (node.arguments.length !== 3) {
```
gitdallas, line 26: I don't think it would get to this point with this being false, unless I am missing something.

packages/eslint-plugin-pf-codemods/lib/rules/v5/resizeObserver-function-param.js
```diff
@@ -0,0 +1,53 @@
...
+                          fixer.insertTextBefore(lastToken, ", false"),
+                        ];
+                        return fixes;
+                      }
+                    },
+                  },
+                  {
+                    desc: "Omit the third argument or pass in a value of true if the callback passed in is not debounced. This is the new default functionality.",
+                    fix: function (fixer) {
+                      if (node.arguments.length !== 3) {
```
gitdallas, line 38: same as above, always true at this point?

packages/eslint-plugin-pf-codemods/lib/rules/v5/resizeObserver-function-param.js
```diff
@@ -0,0 +1,53 @@
...
+                        const lastToken = context.getLastToken(node);
+                        const fixes = [
+                          fixer.insertTextBefore(lastToken, ", false"),
+                        ];
+                        return fixes;
+                      }
+                    },
+                  },
+                  {
+                    desc: "Omit the third argument or pass in a value of true if the callback passed in is not debounced. This is the new default functionality.",
```
gitdallas, line 36: Wouldn't omitting the 3rd argument be the same as passing false?

image

gitdallas (changes requested):
lgtm, just a couple nits and a question on a desc

packages/eslint-plugin-pf-codemods/lib/rules/v5/resizeObserver-function-param.js
thatblindgeye: Good catch! I got the functionality on the v5 branch mixed up with the v4 functionality I think.

thatblindgeye:
@gitdallas thanks for the comments! Made updates and also added some verbiage to the README.

gitdallas (approved):

PR #180, opened 2022-12-16
feat(Popover): remove deprecated props and params

### LINKED ISSUE #140: Popover - Remove deprecated props
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/8201

- Any consumer references to Popover's `boundary` and `tippyProps` props should be removed.
- Any consumers defining and passing Popover a `shouldClose` function with more than one parameter, needs to be given a warning that the first parameter has been removed (or we can remove it for them if possible).
- Any consumers defining parameters as part of functions passed to the following Popover props should be warned that those function types have been updated to remove all parameters:
  - onHidden
  - onHide
  - onMount
  - onShow
  - onShown

_Required actions:_
1. Build codemod
2. Build test

kmcfaul:
Closes #140

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/popover-remove-props.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/popover-remove-props.js
@@ -13,155 +13,61 @@ module.exports = {
     return !popoverImport ? {} : {
       JSXElement(node) {
         if (popoverImport.local.name === node.openingElement.name.name) {
-          const boundaryAttr = node.openingElement.attributes.find(attribute => attribute.name.name === 'boundary');
-          const tippyAttr = node.openingElement.attributes.find(attribute => attribute.name.name === 'tippyProps');
+          const removeProps = ['boundary', 'tippyProps'];
           const shouldCloseAttr = node.openingElement.attributes.find(attribute => attribute.name.name === 'shouldClose');
-          const onHiddenAttr = node.openingElement.attributes.find(attribute => attribute.name.name === 'onHidden');
-          const onHideAttr = node.openingElement.attributes.find(attribute => attribute.name.name === 'onHide');
-          const onMountAttr = node.openingElement.attributes.find(attribute => attribute.name.name === 'onMount');
-          const onShowAttr = node.openingElement.attributes.find(attribute => attribute.name.name === 'onShow');
-          const onShownAttr = node.openingElement.attributes.find(attribute => attribute.name.name === 'onShown');
-
-
-          if (boundaryAttr) {
-            context.report({
-              node,
-              message: "Popover boundary prop has been removed.",
-              fix(fixer) {
-                const fixes = [];
-                fixes.push(fixer.replaceText(boundaryAttr, ''));
-                return fixes;
-              }
-            });
-          }
-
-          if (tippyAttr) {
-            context.report({
-              node,
-              message: "Popover tippyProps prop has been removed.",
-              fix(fixer) {
-                const fixes = [];
-                fixes.push(fixer.replaceText(tippyAttr, ''));
-                return fixes;
-              }
-            });
-          }
-
+          const removeFunctionParam = ['onHidden', 'onHide', 'onMount', 'onShow', 'onShown'];
+          removeProps.forEach(name => {
+            const attr = node.openingElement.attributes.find(attribute => attribute.name.name === name)
+            if (attr) {
+              context.report({
+                node,
+                message: "Popover " + attr.name.name + " prop has been removed.",
+                fix(fixer) {
+                  const fixes = [];
+                  fixes.push(fixer.replaceText(attr, ''));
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin-pf-codemods/lib/rules/v5/popover-remove-props.js
```diff
@@ -0,0 +1,169 @@
...
+          if(shouldCloseAttr) {
+            const params = shouldCloseAttr.value.expression.params;
+            if(params.length === 2) { // only attempt to remove first param if 2 are passed
+              const firstParam = params[0];
+              context.report({
+                node,
+                message: "Popover shouldClose function's first parameter has been removed.",
+                fix(fixer) {
+                  const fixes = [];
+                  fixes.push(fixer.removeRange([firstParam.range[0], firstParam.range[1] + 1])); // remove first property and following comma
```
gitdallas, line 59:
```suggestion
                  fixes.push(fixer.removeRange([firstParam.range[0], params[1].range[0]])); // remove first property up to start of second
```

I think this way would be more flexible (in case users have weird spacing around a comma for example) and also not leave the white space

packages/eslint-plugin-pf-codemods/lib/rules/v5/popover-remove-props.js
```diff
@@ -0,0 +1,169 @@
...
+                  if (funct.range[0] === firstParam.range[0]) {
+                    fixes.push(fixer.replaceText(firstParam, '()'));
+                  } else  {
+                    fixes.push(fixer.remove(firstParam));
+                  }
+                  return fixes;
+                }
+              });
+            }
+          }
```
gitdallas, line 164: I think much of this could be reused/condensed a bit. Also for shouldClose, if we use the second param start range we can remove some ambiguity and an extra white space.

```suggestion
          const removeProps = ['boundary', 'tippyProps'];
          const shouldCloseAttr = node.openingElement.attributes.find(attribute => attribute.name.name === 'shouldClose');
          const removeFunctionParam = ['onHidden', 'onHide', 'onMount', 'onShow', 'onShown'];

          removeProps.forEach(name => {
            const attr = node.openingElement.attributes.find(attribute => attribute.name.name === name)
            if (attr) {
              context.report({
                node,
                message: "Popover " + attr.name.name + " prop has been removed.",
                fix(fixer) {
                  const fixes = [];
                  fixes.push(fixer.replaceText(attr, ''));
                  return fixes;
                }
              });
            }
          });

          if(shouldCloseAttr) {
            const params = shouldCloseAttr.value.expression.params;
            if(params.length === 2) { // only attempt to remove first param if 2 are passed
              const firstParam = params[0];
              const secondParam = params[1];
              context.report({
                node,
                message: "Popover shouldClose function's first parameter has been removed.",
                fix(fixer) {
                  const fixes = [];
                  fixes.push(fixer.removeRange([firstParam.range[0], secondParam.range[0]])); // remove first property and following comma
                  return fixes;
                }
              });
            }
          }

          removeFunctionParam.forEach( name => {
            const attr = node.openingElement.attributes.find(attribute => attribute.name.name === name)
            if(attr) {
              const funct = attr.value.expression;
              if(funct.params.length === 1) {
                const firstParam = funct.params[0];
                context.report({
                  node,
                  message: "Popover " + attr.name.name + " function's parameter has been removed.",
                  fix(fixer) {
                    const fixes = [];
                    if (funct.range[0] === firstParam.range[0]) {
                      fixes.push(fixer.replaceText(firstParam, '()'));
                    } else  {
                      fixes.push(fixer.remove(firstParam));
                    }
                    return fixes;
                  }
                });
              }
            }
          });
          
```

gitdallas (changes requested):

gitdallas (approved):

thatblindgeye (approved):
Just needs to be rebased but looks good!

gitdallas:
conflicts again 😢

gitdallas:
maybe i'm thinking too carefully about these... but what if the consumer was using the params that are getting removed? they are going to have errors now.

i suppose this would serve as their warning to let them know they need to do something different, but curious if anyone had any thoughts on this. @evwilkin

evwilkin:
@gitdallas I'm not as familiar with Tippy & Popover as others, so @kmcfaul may have more insight on the specifics here.

Generally, our release notes should call out any breaking changes so consumers are aware of what they're taking on with the release, our codemods should automate the changes as seamlessly as we can where possible or log warnings as needed to alert the users to additional actions needed.  If a combination of these (codemod updating their code + additional warning logged for them to manually make additional changes) is the best user experience then we should consider that implementation to both do some of the work for them and notify them of where to follow up.

kmcfaul:
boundary and tippyProps have not done anything for a long time now I believe. For the event handlers, a user would encounter errors upgrading to v5 because the prop order has changed so this codemod will help to fix that for them.

PR #181, opened 2022-12-16
feat(Tooltip): remove deprecated props

### LINKED ISSUE #139: Tooltip - remove usages of deprecated props
Follow up to breaking change PR #https://github.com/patternfly/patternfly-react/pull/8231

Remove any usages of Tooltip's `boundary`, `isAppLauncher`, and `tippyProps` props.
They were not causing any side effects anymore so there should be no other clean up necessary.

_Required actions:_
1. Build codemod
2. Build test

kmcfaul:
Closes #139

### DIFF
--- a/packages/pf-codemods/README.md
+++ b/packages/pf-codemods/README.md
@@ -73,3 +73,21 @@ Out:
 ```jsx
 <ToolbarContent visibility={{ default: "hidden" }} />
 ```
+
+### tooltip-remove-props [(#8231)](https://github.com/patternfly/patternfly-react/pull/8231)
+
+We've removed the `boundary`, `tippyProps`, and `isAppLauncher` properties from Tooltip.
+
+#### Examples
+
+In:
+
+```jsx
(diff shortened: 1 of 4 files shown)


gitdallas (approved):

packages/pf-codemods/README.md
```diff
@@ -57,3 +57,21 @@ Out:
...
+In:
+
+```jsx
+<Tooltip boundary={} tippyProps={} isAppLauncher />
+```
+
+Out:
+
+```jsx
+<Tooltip >
```
jenny-s51, line 76: Should this be
```suggestion
<Tooltip  />
```

gitdallas: you're right, would also probably leave 4 spaces

kmcfaul: I'll update. Prettier is automatically removing the extra spacing so if we want to keep the exact output we'll have to watch it because the file needs to be saved without formatting.

gitdallas (approved):

PR #182, opened 2022-12-16
feat(Spinner): remove isSVG

### LINKED ISSUE #146: Spinner - warn isSVG prop default value now true
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/8183

We changed the default value of the `isSVG` prop from `false` to `true`, and should look to potentially send a warning to users where the `Spinner` is used.

_Required actions:_
1. Build codemod
2. Build test

kmcfaul:
Closes #146

This codemod removes the redundant `isSVG` flag now that the default is true. 

I tried to get the codemod to throw a warning for all `Spinner` imports but I haven't quite figured out how to modify the severity of the message (which are all thrown as errors) so I'm holding off on that for now while I tinker with it more locally.

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/spinner-svg-default.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/spinner-svg-default.js
@@ -1,12 +1,28 @@
-const { renameProp } = require('../../helpers');
+const { getPackageImports } = require('../../helpers');
 
 // https://github.com/patternfly/patternfly-react/pull/8183
 module.exports = {
   meta: { fixable: 'code' },
-  create: renameProp(
-    'Spinner',
-    { isSVG: '' },
-    () => `Spinner isSVG prop default value has changed from false to true.`
-  )
+  create: function(context) {
+    const spinnerImports = getPackageImports(context, '@patternfly/react-core')
+      .filter(specifier => specifier.imported.name == 'Spinner');
+      
+    return spinnerImports.length === 0 ? {} : {
+      JSXOpeningElement(node) {
+        if (spinnerImports.map(imp => imp.local.name).includes(node.name.name)) {
+          const svgAttribute = node.attributes.find(n => n.name && n.name.name === 'isSVG');
+          if (svgAttribute && svgAttribute?.value?.expression?.value !== false) {
+            context.report({
+              node,
+              message: `Spinner isSVG prop default value has changed from false to true.`,
+              fix(fixer) {
+                return fixer.replaceText(svgAttribute, ``);
+              }
+            });
+          }
+        }
+      }
+    };
+  }
 };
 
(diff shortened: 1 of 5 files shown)


gitdallas:
i'm curious if we really just want to remove the property. Could this negatively impact people who are using isSVG=false explicitly already?

thatblindgeye (commented):
In regards to @gitdallas  comment above, maybe rather than removing the prop we just warn consumers that the default value has been updated? Or tweaking the message for the fix to mention something like, "If you are passing false to this prop, you can ignore this rule".

kmcfaul:
Yeah the warning bit is what I'm a bit lost on. I can't figure out how to lessen the severity from a full error. I agree that maybe removing the prop isn't necessary here so I'll take another look and remove that.

gitdallas:
this pr has an example that could help with fixing other scenarios https://github.com/patternfly/pf-codemods/pull/205

`isSVG` || `isSVG={true}` = ''
anything else just leave alone, no `context.report()`

kmcfaul:
Agreed with those rules. Added in some extra tests to ensure that `isSVG={true}` gets removed and `isSVG={false}` remains as is.

packages/eslint-plugin-pf-codemods/lib/rules/v5/spinner-svg-default.js
```diff
@@ -0,0 +1,28 @@
...
+  meta: { fixable: 'code' },
+  create: function(context) {
+    const spinnerImports = getPackageImports(context, '@patternfly/react-core')
+      .filter(specifier => specifier.imported.name == 'Spinner');
+      
+    return spinnerImports.length === 0 ? {} : {
+      JSXOpeningElement(node) {
+        if (spinnerImports.map(imp => imp.local.name).includes(node.name.name)) {
+          const svgAttribute = node.attributes.find(n => n.name && n.name.name === 'isSVG');
+          if (svgAttribute && svgAttribute?.value?.expression?.value !== false) {
```
gitdallas, line 14: double checking svgAttribute

packages/pf-codemods/README.md
```diff
@@ -71,7 +71,7 @@ In:
 Out:
 
 ```jsx
-<Card  />
+<Card />
```
gitdallas, line 74: prettier ran?

packages/pf-codemods/index.js
```diff
@@ -35,7 +35,7 @@ function printResults(engine, results, format) {
   }
 
   // Don't show warnings
-  results.forEach(result => result.messages = result.messages.filter(message => message.severity === 2));
+  // results.forEach(result => result.messages = result.messages.filter(message => message.severity === 2));
```
gitdallas, line 38: did you mean to leave this?

gitdallas (commented):
couple comments

packages/pf-codemods/README.md
kmcfaul: Must have, I'll go back and add spaces

packages/pf-codemods/index.js
kmcfaul: No that was something from when I was trying out warnings. Will add it back.

packages/pf-codemods/README.md
kmcfaul: I'm going to leave in the single quotes changing to double quotes because that's what the JSX should look like

Issue #152, opened 2022-12-07
PageGroup - remove deprecated sticky prop


evwilkin:
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/8220

The sticky deprecated prop has been removed from PageGroup and should be removed from consumers' code.

_Required actions:_
1. Build codemod
2. Build test
3. Update readme with description & example

Issue #151, opened 2022-12-07
PageBreadcrumb - remove deprecated sticky prop


evwilkin:
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/8220

The `sticky` deprecated prop has been removed from `PageBreadcrumb` and should be removed from consumers' code.

_Required actions:_
1. Build codemod
2. Build test
3. Update readme with description & example

PR #185, opened 2022-12-22
feat(page components): remove sticky prop

jenny-s51:
Closes:

- #151 
- #152 
- #153 
- #154


gitdallas (changes requested):
to be full and match what we've done other places, think we should add the valid tests for more than just PageSection, and non-pf import tests.

if you want you could make the tests take less real estate (and quicker to glance at) by mapping `['PageSection', 'PageNavigation', etc..]` into the objects.

thatblindgeye (commented):
Other than the above, the only nit I'd have is placing the addition to the README alphabetically.

thatblindgeye (approved):

gitdallas (approved):

PR #186, opened 2023-01-03
feat(simpleList): rename isCurrent to isActive

### LINKED ISSUE #133: SimpleList: Replace prop isCurrent with isActive
Issue:  https://github.com/patternfly/patternfly-react/pull/8132

`isCurrent` is deprecated and has been replaced with `isActive` - functionality remains the same, only name has changed.

_Required actions:_
1. Build codemod
2. Build test

wise-king-sullyman:
Closes #133

### DIFF
--- a/packages/eslint-plugin-pf-codemods/index.js
+++ b/packages/eslint-plugin-pf-codemods/index.js
@@ -44,11 +44,11 @@ const rules = {
   "chartVoronoiContainer-remove-allowTooltip": require('./lib/rules/v4/chartVoronoiContainer-remove-allowTooltip'),
   "chart-remove-allowZoom": require('./lib/rules/v4/chartVoronoiContainer-remove-allowTooltip'),
   "react-icons-remove-icon": require('./lib/rules/v4/react-icons-remove-icon'),
-  
   "datalist-remove-ondrags": require('./lib/rules/v5/datalist-remove-ondrags'),
   "divider-remove-isVertical": require('./lib/rules/v5/divider-remove-isVertical'),
   "fileUpload-remove-onChange": require('./lib/rules/v5/fileUpload-remove-onChange'),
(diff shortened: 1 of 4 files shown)


gitdallas (approved):
Looks good. It's done the same way in several areas but I'm wondering if we should provide a default message for renameProp. Otherwise only use renameProps (even if only a single rename in it like is done in `chartVoronoiContainer-remove-allowTooltip`) as it already provides default messages. Would remove duplicate or inconsistent verbiage.

Many of the places where singular `renameProp` is used is to provide a custom message on purpose - see `accordion-remove-noBoxShadow` for example. Might make sense to allow a `message` prop (even in `renameProps`) and allow that to override the default if necessary... but always provide the default.

This could all be a separate story though.

wise-king-sullyman:
@gitdallas agree, I think it would be beneficial to do one of those. Not sure what the better approach would be off the top of my head though.

Along a similar line of thinking, while I wrote this I debated throwing together a generator or something to do the standard 1 component, 1 prop name being changed codemod like we have here. I figure this format will be pretty common. WDYT?

gitdallas:
@wise-king-sullyman if i understand you correctly, that already exists. https://github.com/patternfly/pf-codemods/blob/master/generate.js

wise-king-sullyman:
@gitdallas that's similar, but not as specific. I whipped together a version of what I was thinking of here: #190 

It would be specifically for these basic single component 1:1 prop replacements specifically, which seem somewhat common. Though maybe it's a bit *too* specific to be all that useful.

thatblindgeye (commented):
Just some conflicts to resolve but otherwise looks good! Would be worth placing the section in the README alphabetically as part of resolving those

thatblindgeye (approved):

PR #190, opened 2023-01-04
chore(helpers): add prop replacement specific generator

wise-king-sullyman:
Quick PR demonstrating a more niche generator specifically for prop replacement.

I'm not sure if we would want to add this or not.

If we do like this idea of more specific generators I could throw together a few more for common use cases.

Thoughts?


gitdallas:
do you think it would make sense to use the existing generator and add some optional args?

gitdallas (approved):

Issue #147, opened 2022-12-07
Pagination - renamed prop


evwilkin:
Follow up to breaking change PR [[link]](https://github.com/patternfly/patternfly-react/pull/8134)

In the `pagination` component we've renamed a prop from `ToggleTemplateProps` to `PaginationToggleTemplateProps` and should push this change out to consumers' code.

_Required actions:_
1. Build codemod
2. Build test
3. Update readme with description & example

PR #201, opened 2023-01-06
feat(Pagination): replace toggleTemplateProps with paginationToggleTe…

gitdallas:
closes: #147

### DIFF
--- a/generate-replace-prop-v5.js
+++ b/generate-replace-prop-v5.js
@@ -9,7 +9,7 @@ const lowerCasedComponentName = componentName[0].toLowerCase() + componentName.s
 const newRuleName = `${lowerCasedComponentName}-remove-${oldPropName}`
 
 if (!componentName || !oldPropName || !newPropName || !referencePR) {
-  console.log('usage: node generate [componentName] [oldPropName] [newPropName] [referencePR]');
+  console.log('usage: node generate-replace-prop-v5 [componentName] [oldPropName] [newPropName] [referencePR]');
   process.exit(1);
 }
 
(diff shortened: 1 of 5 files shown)


thatblindgeye (approved):
Looks good! Only nit is using `rename` vs `remove`, but not a blocker.

gitdallas:
> Looks good! Only nit is using `rename` vs `remove`, but not a blocker.

Gonna blame @wise-king-sullyman's generator 😜

wise-king-sullyman:
I mean, the old prop _was_ removed was it not 😂 

Really though I'd be on board with standardizing on calling a rename mod a `rename`, and an actual removal a `remove`, it just seemed that we were just naming the rules `remove` for either case so far from what I could see so I went with that.

gitdallas:
@wise-king-sullyman I suppose the generator could check if the replacement is empty str to determine and do both ways

wise-king-sullyman:
Yeah I'm still planning to circle back to the generator when I get a free moment, restructure things a bit so that handle a few other common use cases like that in a more straightforward way.

That would definitely be a quicker way to implement replace and remove functionality though. I might throw up a quick PR for that.

gitdallas:
@wise-king-sullyman but can you pr review? 😅

generate-replace-prop-v5.js
```diff
@@ -9,7 +9,7 @@ const lowerCasedComponentName = componentName[0].toLowerCase() + componentName.s
 const newRuleName = `${lowerCasedComponentName}-remove-${oldPropName}`
 
 if (!componentName || !oldPropName || !newPropName || !referencePR) {
-  console.log('usage: node generate [componentName] [oldPropName] [newPropName] [referencePR]');
+  console.log('usage: node generate-replace-prop-v5 [componentName] [oldPropName] [newPropName] [referencePR]');
```
wise-king-sullyman, line 12: Whoops, nice catch.

wise-king-sullyman (approved):
Yes

PR #205, opened 2023-01-10
feat(Tabs): rename tabs props

### LINKED ISSUE #203: Tabs - border bottom prop updates
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/8517/files

1. Rename `hasBorderBottom` to `hasNoBorderBottom`
2. Delete `hasSecondaryBorderBottom`

_Required actions:_
1. Build codemod
3. Build test
4. Update readme with description & example

jenny-s51:
Closes #203

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/tabs-rename-hasBorderBottom.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/tabs-rename-hasBorderBottom.js
@@ -1,0 +1,32 @@
+const { getPackageImports } = require('../../helpers');
+
+// https://github.com/patternfly/patternfly-react/pull/8517
+module.exports = {
+  meta: { fixable: 'code' },
+  create: function(context) {
+    const imports = getPackageImports(context, '@patternfly/react-core')
+      .filter(specifier => specifier.imported.name === 'Tabs');
+    const sourceCode = context.getSourceCode();
+      
+    return imports.length == 0 ? {} : {
+      JSXOpeningElement(node) {
+        if (imports.map(imp => imp.local.name).includes(node.name.name)) {
+          const attribute = node.attributes.find(node => node.name && node.name.name === 'hasBorderBottom');
+          if (attribute) {
+            context.report({
+              node,
+              message: `hasBorderBottom prop has been removed for ${node.name.name}. Use hasNoBorderBottom instead`,
+              fix(fixer) {
+                const attrText = attribute.value ? sourceCode.getText(attribute.value.expression) : 'true';
+                if (['true', 'false'].includes(attrText.trim())) {
+                  return fixer.replaceText(attribute, attrText && attrText.trim() === 'false' ? 'hasNoBorderBottom' : '');
+                }
+                return fixer.replaceText(attribute, `hasNoBorderBottom={!(${attrText})}`)
+              }
+            });
+          }
+        }
+      }
+    };
+  }
+};
--- a/packages/pf-codemods/README.md
+++ b/packages/pf-codemods/README.md
@@ -368,6 +368,48 @@ Out:
 <Spinner  />
 ```
 
+### tabs-rename-hasBorderBottom [(#8517)](https://github.com/patternfly/patternfly-react/pull/8517)
+
+We've renamed the `hasBorderBottom` prop to `hasNoBorderBottom`.
+
+#### Examples
+
+In:
+
+```jsx
+<Tabs hasBorderBottom />
+<Tabs hasBorderBottom={true} />
+<Tabs hasBorderBottom={false} />
+<Tabs hasBorderBottom={someVar} />
+```
+
+Out:
+
+```jsx
+<Tabs  />
+<Tabs  />
+<Tabs hasNoBorderBottom />
+<Tabs hasNoBorderBottom={!someVar} />
+```
+
+### tabs-rename-hasSecondaryBorderBottom [(#8517)](https://github.com/patternfly/patternfly-react/pull/8517)
+
+We've removed the deprecated `hasSecondaryBorderBottom` prop.
+
+#### Examples
+
+In:
+
+```jsx
+<Tabs hasSecondaryBorderBottom />
+```
+
+Out:
+
+```jsx
+<Tabs  />
+```
+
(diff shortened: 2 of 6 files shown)


packages/pf-codemods/README.md
```diff
@@ -249,6 +249,25 @@ Out:
...
+### tabs-rename-props [(#8212)](https://github.com/patternfly/patternfly-react/pull/8517)
+
+We've removed the deprecated `hasSecondaryBorderBottom` prop, and renamed the `hasBorderBottom` prop to `hasNoBorderBottom`.
+
+#### Examples
+
+In:
+
+```jsx
+<Tabs hasNoBorderBottom hasSecondaryBorderBottom />
```
gitdallas, line 261:
```suggestion
<Tabs hasBorderBottom hasSecondaryBorderBottom />
```

packages/pf-codemods/README.md
```diff
@@ -249,6 +249,25 @@ Out:
 <SimpleList isActive />
 ```
 
+### tabs-rename-props [(#8212)](https://github.com/patternfly/patternfly-react/pull/8517)
```
gitdallas, line 252:
```suggestion
### tabs-rename-props [(#8517)](https://github.com/patternfly/patternfly-react/pull/8517)
```

packages/eslint-plugin-pf-codemods/test/rules/v5/tabs-rename-props.js
```diff
@@ -0,0 +1,36 @@
+const ruleTester = require("../../ruletester");
+const rule = require("../../../lib/rules/v5/tabs-rename-props");
+
+ruleTester.run("tabs-rename-props", rule, {
+  valid: [
+    {
+      code: `import { Tabs } from '@patternfly/react-core'; <Tabs />`,
```
gitdallas, line 7:
```suggestion
      code: `import { Tabs } from '@patternfly/react-core'; <Tabs hasNoBorderBottom />`,
```

gitdallas (changes requested):
just a couple little things

packages/eslint-plugin-pf-codemods/lib/rules/v5/tabs-rename-props.js
```diff
@@ -0,0 +1,14 @@
+const { renameProps } = require("../../helpers");
+
+const renames = {
+  Tabs: {
+    hasSecondaryBorderBottom: "",
+    hasBorderBottom: "hasNoBorderBottom",
```
thatblindgeye, line 6: Should this codemod just rename the hasBorderBottom prop? @gitdallas would this be similar to your comment on the isSVG PR? Just wondering if simply renaming it might cause issues i.e. if someone has passed in `hasBorderBottom={false}` to remove the borders, would this then update things so that it becomes `hasNoBorderBottom={false}` and have borders rendered.

gitdallas: good point, the boolean value should be flipped

gitdallas: good catch/point, the boolean value should be flipped. I think we can still do a code fix like below, wdyt? @thatblindgeye 

`hasBorderBottom` => ""
`hasBorderBottom={true}` => ""
`hasBorderBottom={false}` => `hasNoBorderBottom`
`hasBorderBottom={someVar}` => `hasNoBorderBottom{!someVar}`

thatblindgeye: That sounds good to me! I imagine that'd cover the basic use cases of it.

gitdallas: started a slack convo regarding these more complex cases (here)[https://patternfly.slack.com/archives/C048DB6UR2P/p1673450939845399]... but for example, the `someVar` could also be more complex so perhaps may need to be wrapped in `()` in case it is something like `someVar || someOtherVar` ... also trying to get a feel for if we should even be taking on these complex situations vs alerting the consumer and asking them to do it themselves.

jenny-s51 (commented):
Great catch @thatblindgeye, this is certainly a good start to a larger discussion regarding other instances like this @gitdallas 

I've updated this PR with changes that [emulate what we did for the renaming of `hasBodyPadding` within the v4 Wizard component](https://github.com/patternfly/pf-codemods/blob/1402d45e1367f745db12dfae95aa00a5add3f864/packages/eslint-plugin-pf-codemods/lib/rules/v4/wizard-rename-hasBodyPadding.js#L4).

packages/eslint-plugin-pf-codemods/lib/rules/v5/tabs-rename-hasBorderBottom.js
```diff
@@ -0,0 +1,29 @@
...
+      JSXOpeningElement(node) {
+        if (imports.map(imp => imp.local.name).includes(node.name.name)) {
+          const attribute = node.attributes.find(node => node.name && node.name.name === 'hasBorderBottom');
+          if (attribute) {
+            context.report({
+              node,
+              message: `hasBorderBottom prop has been removed for ${node.name.name}. Use hasNoBorderBottom instead`,
+              fix(fixer) {
+                const attrText = attribute.value ? sourceCode.getText(attribute.value.expression) : null;
+                return fixer.replaceText(attribute, attrText && attrText.includes('false') ? 'hasNoBorderBottom' : '');
```
gitdallas, line 21:
```suggestion
                const attrText = attribute.value ? sourceCode.getText(attribute.value.expression) : 'true';
                if (['true', 'false'].includes(attrText)) {
                  return fixer.replaceText(attribute, attrText && attrText === 'false' ? 'hasNoBorderBottom' : '');
                }
                return fixer.replaceText(attribute, `hasNoBorderBottom={!(${attrText})}`)
```

gitdallas (commented):
this looks good and i've learned some from it. i wonder if we should also consider they might pass a variable or expression other than just `true` or `false`.

do you think the below change would handle that well?

gitdallas (approved):

thatblindgeye (approved):
🎉 Looks nice! Just needs the conflicts resolved again

wise-king-sullyman (approved):

PR #206, opened 2023-01-11
Charts - replace getResizeObserver

dlabrecq:
Replaced react-chart `getResizeObserver` with same function from react-core.

Related to PR https://github.com/patternfly/patternfly-react/pull/8533

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/charts-resizeObserver-import.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/charts-resizeObserver-import.js
@@ -1,0 +1,54 @@
+const { getPackageImports } = require("../../helpers");
+
+// https://github.com/patternfly/patternfly-react/pull/8533
+module.exports = {
+  meta: { fixable: "code" },
+  create: function (context) {
+    const chartResizeImport = getPackageImports(
+      context,
+      "@patternfly/react-charts"
+    ).filter((specifier) => specifier.imported.name == "getResizeObserver");
+
+    return chartResizeImport.length === 0
+      ? {}
+      : {
+          ImportDeclaration(node) {
+            if (node.source.value === "@patternfly/react-charts") {
+              context.report({
+                node,
+                message: `The getResizeObserver function has been removed from react-charts and should be imported from react-core instead.`,
+                fix(fixer) {
+                  const fixes = [];
+                  if (node.specifiers.length > 1) {
+                    const { range } = chartResizeImport[0];
+                    const nextToken = context
+                      .getSourceCode()
+                      .getTokenAfter(chartResizeImport[0]);
+                    const rangeEnd =
+                      nextToken && nextToken.value === ","
+                        ? range[1] + 1
+                        : range[1];
+
+                    fixes.push(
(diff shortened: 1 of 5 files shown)


nicolethoen (approved):

packages/pf-codemods/README.md
```diff
@@ -324,3 +324,21 @@ Out:
 ```jsx
 <Tooltip     />
 ```
+
+### chart-resizeObserver-function [(#8533)](https://github.com/patternfly/patternfly-react/pull/8533)
```
thatblindgeye, line 328: Nit: can this be placed alphabetically in the README

thatblindgeye (changes requested):
If there's a likelihood that react-charts' getResizeObserver was being imported by anyone else, and since the README includes an "Examples" section for in and output, it might be worth including a codemod that can be ran to fix the import from `'@patternfly/react-charts'` to `@patternfly/react-core'`. 

Alternatively we could just mention the change and remove the Examples section in the README. @gitdallas wdyt?

gitdallas:
@thatblindgeye - i believe nicole just let dan know he should either create a story or at least update the readme for this. maybe one of us could take over creating the codemod for this - unless you want to, @dlabrecq

packages/pf-codemods/README.md
dlabrecq: Done

dlabrecq:
Per Nicole's instructions, I created issue https://github.com/patternfly/patternfly-react/pull/8533 and updated the README in this repo. I wasn't planning to create and test codemods, so I'm fine having someone else implement that.

thatblindgeye (commented):
Sounds good! @gitdallas unless you'd want to add a codemod and tests for this, I could push those changes to this PR

thatblindgeye:
@gitdallas let me know if you think the codemod should be simpler and just update the import path. Tried taking into account other scenarios with what I pushed.

packages/eslint-plugin-pf-codemods/lib/rules/v5/charts-resizeObserver-import.js
```diff
@@ -0,0 +1,54 @@
...
+                        ? range[1] + 1
+                        : range[1];
+
+                    fixes.push(
+                      fixer.replaceTextRange([range[0], rangeEnd], "")
+                    );
+                    fixes.push(
+                      fixer.insertTextAfterRange(
+                        node.range,
+                        "import {getResizeObserver} from '@patternfly/react-core';"
```
gitdallas, line 38: should we sneak a `\n` in before "import"?

gitdallas:
thanks, lgtm. while you resolve - this would be a good opportunity to add to the `test.tsx` file btw and see the output with `yarn test:single`

gitdallas (approved):

thatblindgeye:
@gitdallas tried out the test.tsx file, pretty neat! Wasn't sure if we should keep additions saved as part of PRs or if it's better just to try locally so I didn't include it here.

dlabrecq:
@thatblindgeye Thank you for adding the codemods.

PR #207, opened 2023-01-13
feat(ClipboardCopy): removed PopoverPosition type

### LINKED ISSUE #160: ClipboardCopy & ClipboardCopyButton - updated position type
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/8226

The `position` prop in both the `ClipboardCopy` and `ClipboardCopyButton` components has an updated typing in that in no longer includes `PopoverPosition`.  TBD:  is consumer code change needed, or simply a warning to indicate this typing change?

_Required actions:_
1. Build codemod
2. Build test

thatblindgeye:
Closes #160

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/clipboardCopy-remove-popoverPosition.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/clipboardCopy-remove-popoverPosition.js
@@ -24,8 +24,8 @@ module.exports = {
                 (nodeAttr) => nodeAttr.name && nodeAttr.name.name === "position"
               );
               if (
-                positionAttr.value.expression &&
-                positionAttr.value.expression.object.name === "PopoverPosition"
+                positionAttr?.value?.expression?.object?.name ===
+                "PopoverPosition"
               ) {
                 context.report({
                   node,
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin-pf-codemods/lib/rules/v5/clipboardCopy-remove-popoverPosition.js
```diff
@@ -0,0 +1,45 @@
...
+              clipboardImports
+                .map((imp) => imp.local.name)
+                .includes(node.name.name)
+            ) {
+              const positionAttr = node.attributes.find(
+                (nodeAttr) => nodeAttr.name && nodeAttr.name.name === "position"
+              );
+              if (
+                positionAttr.value.expression &&
+                positionAttr.value.expression.object.name === "PopoverPosition"
```
gitdallas, line 28: `positionAttr` could be undefined at this point which would cause an error, think you could just..
```suggestion
                positionAttr?.value?.expression?.object?.name === "PopoverPosition"
```

gitdallas (changes requested):
This may result in some unused imports.. but I suppose it would probably be too much to try checking if that import is still being used elsewhere and a regular lint could catch that after so should be okay.

Did see a possible NPE though:

packages/eslint-plugin-pf-codemods/lib/rules/v5/clipboardCopy-remove-popoverPosition.js
gitdallas: can test this with a `valid` test like:

    code: `import { ClipboardCopy } from '@patternfly/react-core'; <ClipboardCopy />`

gitdallas (approved):

PR #208, opened 2023-01-16
feat(Pagination): add codemod for optionstoggle

### LINKED ISSUE #162: Pagination - OptionsToggle replaced with Menu/MenuToggle
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/8319

Likely, the best a code mod can do here is warn any consumers importing or referencing Pagination's OptionsToggle, that it has been removed. @thatblindgeye is there anything else?

_Required actions:_
1. Build codemod
2. Build test

kmcfaul:
Closes #162

Doesn't have a fixer because replacing the component would be more complex than a 1:1 replacement, and also because it's unlikely this component is being used directly. If we figure out how to set the severity level this would be a good one to update to warn instead.


gitdallas (approved):

kmcfaul:
Might be better to leave as an error because a direct import would cause issues.

wise-king-sullyman (approved):

PR #209, opened 2023-01-16
feat(Nav): updated aria-label default value

### LINKED ISSUE #141: Nav - horizontalsubnav aria label default updated
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/8213

Any consumer using the nav with a variant `horizontal-subnav` should be warned that the default aria label has been updated from 'Global' to 'Local'

_Required actions:_
1. Build codemod
2. Build test

thatblindgeye:
Closes #141

### DIFF
--- a/packages/eslint-plugin-pf-codemods/index.js
+++ b/packages/eslint-plugin-pf-codemods/index.js
@@ -53,6 +53,7 @@ const rules = {
   "datalist-remove-ondrags": require('./lib/rules/v5/datalist-remove-ondrags'),
   "divider-remove-isVertical": require('./lib/rules/v5/divider-remove-isVertical'),
   "fileUpload-remove-onChange": require('./lib/rules/v5/fileUpload-remove-onChange'),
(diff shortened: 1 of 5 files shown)


gitdallas (approved):
lgtm, bonus could be checking if aria-label is defined and skipping the warning if so

@wise-king-sullyman review for potential "warning" type?

thatblindgeye:
>lgtm, bonus could be checking if aria-label is defined and skipping the error if so

@gitdallas good call, added that in!

wise-king-sullyman (approved):
LGTM as well, though do you think this is something that would make more sense as a warning? If I understand things right this wouldn't cause any actual errors to be thrown if consumers don't update it.

gitdallas:
only errors might be in tests. if anything is a warning this would be it.. have to get #213  in first then

gitdallas:
@thatblindgeye - #213 is in. if you agree with this being a warning and wouldn't mind adding this to the warning array while resolving

thatblindgeye:
@gitdallas @wise-king-sullyman added to the warning array!

PR #210, opened 2023-01-16
feat(KEY_CODES): warn about KEY_CODES removal

### LINKED ISSUE #134: Warning - KEY_CODES removed
Original issue: https://github.com/patternfly/patternfly-react/pull/8174

We have removed exported const `KEY_CODES`, which was not used anywhere within our code base but we should warn any users importing this that is has been removed.

_Required actions:_
1. Build codemod
2. Build test

wise-king-sullyman:
Closes #134 

Also address some bugs and enables us to have rules as warnings rather than just errors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,10 +7,10 @@
     "test": "mocha packages/eslint-plugin-pf-codemods/test/rules/*",
     "test:v4": "mocha packages/eslint-plugin-pf-codemods/test/rules/v4",
     "test:v5": "mocha packages/eslint-plugin-pf-codemods/test/rules/v5",
-    "test:single": "pf-codemods test/test.tsx",
-    "test:koku": "pf-codemods test/koku-ui",
-    "test:console": "pf-codemods test/console/frontend",
-    "test:integreatly": "pf-codemods test/tutorial-web-app"
+    "test:single": "pf-codemods --no-cache test/test.tsx",
+    "test:koku": "pf-codemods --no-cache test/koku-ui",
+    "test:console": "pf-codemods --no-cache test/console/frontend",
+    "test:integreatly": "pf-codemods --no-cache test/tutorial-web-app"
   },
   "repository": {
     "type": "git",
--- a/packages/eslint-plugin-pf-codemods/index.js
+++ b/packages/eslint-plugin-pf-codemods/index.js
@@ -62,12 +62,13 @@ const rules = {
   "tooltip-remove-props": require('./lib/rules/v5/tooltip-remove-props'),
   "remove-removeFindDomNode": require('./lib/rules/v5/remove-removeFindDomNode'),
   "popover-remove-props": require('./lib/rules/v5/popover-remove-props'),
+  "warn-key-codes-removed": require('./lib/rules/v5/warn-key-codes-removed'),
 };
 
 module.exports = {
   configs: {
     recommended: {
-      plugins: ["pf-codemods"],
+      plugins: ["@patternfly/pf-codemods"],
       env: {
         browser: true,
         node: true,
@@ -83,7 +84,8 @@ module.exports = {
         },
       },
       rules: Object.keys(rules).reduce((acc, rule) => {
-        acc[`pf-codemods/${rule}`] = "error";
+        const severity = rule.includes('warn') ? 'warn' : 'error';
+        acc[`@patternfly/pf-codemods/${rule}`] = severity;
         return acc;
       }, {}),
     },
--- a/packages/eslint-plugin-pf-codemods/index.js
+++ b/packages/eslint-plugin-pf-codemods/index.js
@@ -84,8 +84,7 @@ module.exports = {
         },
       },
       rules: Object.keys(rules).reduce((acc, rule) => {
-        const severity = rule.includes('warn') ? 'warn' : 'error';
-        acc[`@patternfly/pf-codemods/${rule}`] = severity;
+        acc[`@patternfly/pf-codemods/${rule}`] = "error";
         return acc;
       }, {}),
     },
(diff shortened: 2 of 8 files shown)


packages/eslint-plugin-pf-codemods/index.js
```diff
@@ -62,12 +62,13 @@ const rules = {
...
   "remove-removeFindDomNode": require('./lib/rules/v5/remove-removeFindDomNode'),
   "popover-remove-props": require('./lib/rules/v5/popover-remove-props'),
+  "warn-key-codes-removed": require('./lib/rules/v5/warn-key-codes-removed'),
 };
 
 module.exports = {
   configs: {
     recommended: {
-      plugins: ["pf-codemods"],
+      plugins: ["@patternfly/pf-codemods"],
```
wise-king-sullyman, line 71: This fixes an error introduced when we scoped the packages under `@patternfly`

packages/eslint-plugin-pf-codemods/index.js
```diff
@@ -83,7 +84,8 @@ module.exports = {
         },
       },
       rules: Object.keys(rules).reduce((acc, rule) => {
-        acc[`pf-codemods/${rule}`] = "error";
+        const severity = rule.includes('warn') ? 'warn' : 'error';
+        acc[`@patternfly/pf-codemods/${rule}`] = severity;
```
wise-king-sullyman, line 88: This allows us to set rules as warnings on the `eslint-plugin-pf-codemods` side of things.

Alternatively we could set it in the `eslintrc` file in `pf-codemods`, not sure which approach would be better and fine with either, but I lean *slightly* towards this so that it's a warning regardless of if the `pf-codemods` package is being used or the `eslint-plugin-pf-codemods`.

packages/pf-codemods/index.js
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 const fspath = require('path');
 const { CLIEngine } = require('eslint/lib/cli-engine');
-const { configs } = require('eslint-plugin-pf-codemods');
+const { configs } = require('@patternfly/eslint-plugin-pf-codemods');
```
wise-king-sullyman, line 4: Another fix for the move to the `@patternfly` scope.

packages/pf-codemods/index.js
```diff
@@ -34,9 +34,6 @@ function printResults(engine, results, format) {
     return false;
   }
 
-  // Don't show warnings
-  results.forEach(result => result.messages = result.messages.filter(message => message.severity === 2));
-
```
wise-king-sullyman, line 39: This setting was making it so that warnings never surfaced before, it has to be removed so that we can issue them.

gitdallas: can you think of a reason this was here in the first place?

packages/eslint-plugin-pf-codemods/lib/rules/v5/warn-key-codes-removed.js
```diff
@@ -0,0 +1,23 @@
+const { getPackageImports } = require("../../helpers");
+
+// https://github.com/patternfly/patternfly-react/pull/8174
+module.exports = {
+  meta: { fixable: "code" },
```
gitdallas, line 5: should this be here if we aren't doing a code fix? also, did you mean this rule to add a warning vs error?

packages/pf-codemods/index.js
wise-king-sullyman: Not off the top of my head, and it was just added in this commit that changed things all over the repo https://github.com/patternfly/pf-codemods/commit/1ab03a0f18667c74ff23812791abb81c05ca6932#diff-06072659dcb0ba81f72d49d18cc77c030ba13d5edacde533742c0dd8ac81730f

packages/eslint-plugin-pf-codemods/lib/rules/v5/warn-key-codes-removed.js
wise-king-sullyman: Hmmm, probably not? I'll try removing and see what happens.

And yes, the intent was for it to be a warning. If we want it to be an error though I'm fine with that.

gitdallas (approved):

packages/eslint-plugin-pf-codemods/index.js
wise-king-sullyman: I've since removed this as it seems some more discussion probably needs to happen around how we'd want this implemented, or if we even actually want it implemented.

package.json
```diff
@@ -7,10 +7,10 @@
...
     "test:v4": "mocha packages/eslint-plugin-pf-codemods/test/rules/v4",
     "test:v5": "mocha packages/eslint-plugin-pf-codemods/test/rules/v5",
-    "test:single": "pf-codemods test/test.tsx",
-    "test:koku": "pf-codemods test/koku-ui",
-    "test:console": "pf-codemods test/console/frontend",
-    "test:integreatly": "pf-codemods test/tutorial-web-app"
+    "test:single": "pf-codemods --no-cache test/test.tsx",
+    "test:koku": "pf-codemods --no-cache test/koku-ui",
+    "test:console": "pf-codemods --no-cache test/console/frontend",
+    "test:integreatly": "pf-codemods --no-cache test/tutorial-web-app"
```
wise-king-sullyman, line 13: This makes it so that tests don't create a cache file, the cache file was causing issues with testing and needing to be manually deleted after each run.

packages/pf-codemods/index.js
```diff
@@ -13,6 +13,7 @@ program
   .option('--exclude <rules>', 'Run recommended rules EXCLUDING this comma-seperated list')
   .option('--fix', 'Whether to run fixer')
   .option('--format <format>', 'What eslint report format to use', 'stylish')
+  .option('--no-cache', 'Disables eslint caching')
```
wise-king-sullyman, line 16: This + line 76 add a CLI option that allows us to selectively use eslint caching. Defaults to true so that consumers using it still get the benefits of the caching.

You can read more about what I'm actually doing here if you're curious: https://www.npmjs.com/package/commander#options

thatblindgeye (approved):

PR #213, opened 2023-01-17
feat(rules): add severity customization

### LINKED ISSUE #211: Enable setting the severity level of rules
Follow up to #210. Moving to its own issue rather than trying to tackle in #210 so that #210 isn't blocked by it.

We should maybe figure out a good way to enable severity level changes in our rules. 

_Required actions:_
1. Determine if we want to have severity options or not
1. Figure out the best way to implement severity customization if we want it
1. Enable the severity customization

wise-king-sullyman:
Closes #211


gitdallas (approved):
lgtm, might be worth adding a note in the contributing readme

thatblindgeye (approved):

PR #214, opened 2023-01-18
separate v4 rules

gitdallas:
with this we'll be able to add the `--v4` flag to run the v4 codemods. also split up the `test.tsx` and `test:single` into v4 and v5


wise-king-sullyman (approved):
I like it!

wise-king-sullyman (approved):

nicolethoen (approved):
LGTM!

My only thought is whether the changes to the pf-codemods/index.js are going to be easily adjusted for future major releases with the v4/v5 hard coded in so many places.

Issue #142, opened 2022-11-23
FileUpload - Types changed after react-dropzone upgrade


nicolethoen:
Follow up to breaking change PR [link]

I'm not positive if we can do anything here other than warn consumers of FileUpload that props they are using have had types updated.

The prop types of FileUpload's `onFileInputChange` prop and `dropzoneProps` prop have been updated.
The prop types of MultipleFileUpload's `dropzoneProps` prop has been updated.

_Required actions:_
1. Build codemod
2. Build test
3. Update readme with description & example

PR #216, opened 2023-01-20
feat(react-dropzone): updated to latest version

### LINKED ISSUE #166: React-dropzone - warn consumers that major version has been updated
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/7926

This may not be something we can write a code mod for - some products might be relying (unadvisedly) on the PF react-dropzone dependency, so they'd need to make similar breaking changes in their project. We could print out a warning if they are using react-dropzone in their product at all?

_Required actions:_
1. Build codemod
2. Build test

thatblindgeye:
Closes #166

### DIFF
--- a/packages/pf-codemods/README.md
+++ b/packages/pf-codemods/README.md
@@ -386,6 +386,10 @@ Out:
 />
 ```
 
+### react-dropzone-warn-upgrade [(#7926)](// https://github.com/patternfly/patternfly-react/pull/7926)
+
+The `react-dropzone` dependency used with FileUpload, MultipleFileUpload, and CodeEditor has been updated from version 9 to version 14.
+
 ### remove-removeFindDomNode [(#8371)](https://github.com/patternfly/patternfly-react/pull/8371) [(#8316)](https://github.com/patternfly/patternfly-react/pull/8316)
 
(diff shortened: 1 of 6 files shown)


packages/pf-codemods/README.md
```diff
@@ -294,6 +294,10 @@ Out:
 />
 ```
 
+### react-dropzone-warn-upgrade [(#7926)](// https://github.com/patternfly/patternfly-react/pull/7926)
+
+The `react-dropzone` dependency used with FileUpload, MultipleFileUpload, and CodeEditor has been updated from version 9 to version 14.
```
gitdallas, line 299: What might a consumer need to do with this change?

thatblindgeye: From the PR itself, other than updating the version it looks like there were some type updates. Other than possibly mentioning those in the warning message, we could just state that the version bump may require additional breaking changes to be made (though it's a bit generic)

thatblindgeye: Though part of that may be resolved with https://github.com/patternfly/pf-codemods/issues/142

gitdallas (approved):

packages/pf-codemods/README.md
wise-king-sullyman: Hm, I wonder if this and a PR for #142 are both needed 🤔

gitdallas: right... i wonder if the language from this can be merged in (if we still want to talk about version numbers)

wise-king-sullyman: Yeah that would make sense to me. Want me to do that under a followup?

PR #220, opened 2023-01-20
feat(Tabs): added warning that the type of Tabs children is now limited

### LINKED ISSUE #168: Tabs - children prop type updated
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/8217

Anyone using the Tabs component, we need to warn them that the type of the children prop has been updated from `React.ReactNode` (which was documented to need to be actually `Must be React.ReactElement<TabProps>[]`) to now be `TabsChild | TabsChild[]` (which is now documented to mean the same as `Only 'Tab' components or expressions resulting in a falsy value`)

_Required actions:_
1. Build codemod
2. Build test

wise-king-sullyman:
Closes #168

### DIFF
--- a/generate.js
+++ b/generate.js
@@ -69,12 +69,3 @@ ruleTester.run("${newRuleName}", rule, {
 });
 `
 );
-
-// Update index file
-const ruleIndexPath = path.join(__dirname, 'packages/eslint-plugin-pf-codemods/index.js');
-const ruleIndex = fs.readFileSync(ruleIndexPath, 'utf8');
-fs.writeFileSync(
-  ruleIndexPath,
-  // (ab)Use fact that `rules` object is at top of file
-  ruleIndex.replace("};", `  "${newRuleName}": require('./lib/rules/v5/${newRuleName}'),\n};`)
-);
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/tabs-warn-children-type-changed.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/tabs-warn-children-type-changed.js
@@ -1,0 +1,23 @@
+const { getPackageImports } = require("../../helpers");
+
+// https://github.com/patternfly/patternfly-react/pull/8217
+module.exports = {
+  create: function (context) {
+    return {
+      ImportDeclaration(node) {
+        const TabsImport = node.specifiers.find(
+          (specifier) =>
+            specifier.imported.name === "Tabs" &&
+            node.source.value === "@patternfly/react-core"
+        );
+
+        if (TabsImport) {
+          context.report({
+            node,
+            message: "The children of the 'Tabs' component must now be passed a 'Tab' component or a falsy value.",
+          });
+        }
+      },
+    };
+  },
+};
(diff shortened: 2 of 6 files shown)


generate.js
```diff
@@ -69,12 +69,3 @@ ruleTester.run("${newRuleName}", rule, {
...
 );
-
-// Update index file
-const ruleIndexPath = path.join(__dirname, 'packages/eslint-plugin-pf-codemods/index.js');
-const ruleIndex = fs.readFileSync(ruleIndexPath, 'utf8');
-fs.writeFileSync(
-  ruleIndexPath,
-  // (ab)Use fact that `rules` object is at top of file
-  ruleIndex.replace("};", `  "${newRuleName}": require('./lib/rules/v5/${newRuleName}'),\n};`)
-);
```
wise-king-sullyman, line 80: Removed as the index file no longer needs to be manually updated with rule names.

packages/eslint-plugin-pf-codemods/lib/rules/v5/tabs-warn-children-type-changed.js
```diff
@@ -0,0 +1,23 @@
...
+        const TabsImport = node.specifiers.find(
+          (specifier) =>
+            specifier.imported.name === "Tabs" &&
+            node.source.value === "@patternfly/react-core"
+        );
+
+        if (TabsImport) {
+          context.report({
+            node,
+            message: "Children of the 'Tabs' component must now be 'Tab' components or expressions resulting in a falsy value.",
```
wise-king-sullyman, line 17: I don't _love_ this message, but it was the best I could come up with. 100% open to suggestions.

gitdallas: idk why but when i see "children of the" i think "corn" 😆 

how about "`Tabs` component children must be `Tab` components"

not sure that the falsy needs to be addressed. technically it would be "expressions resulting in either falsy value or a Tab component"

thatblindgeye: "The `children` prop of the Tabs component must now be passed a Tab component or a falsy value."?

wise-king-sullyman: @thatblindgeye  I like the sound of that, but since `children` isn't usually used explicitly as a prop I wonder if that might trip people up 🤔 

@gitdallas so your suggestion would be "Tabs component children must be Tab components or expressions resulting in either falsy value or a Tab component" in its full form?

gitdallas: @wise-king-sullyman - i really don't think the mention of expressions is necessary, anything could be expression but that doesn't matter... all that matters is what comes out of the expression. 

i was thinking "`Tabs` component children must be `Tab` components or a falsy value", I guess... or in Eric's format we could have "The `children` of the `Tabs` component must now be passed a `Tab` component or a falsy value."

packages/eslint-plugin-pf-codemods/lib/rules/v5/tabs-warn-children-type-changed.js
```diff
@@ -0,0 +1,23 @@
+const { getPackageImports } = require("../../helpers");
```
thatblindgeye, line 1: Nit: could remove this

wise-king-sullyman: Good catch!

gitdallas (approved):

Issue #225, opened 2023-01-25
Code editor - Remove deprecated props


tlabaj:
Follow up to breaking change issue https://github.com/patternfly/patternfly-react/issues/8214

the following props have been deprecated and should be removed
entryDelay
exitDelay
maxWidth
position
tooltiptext

_Required actions:_
1. Build codemod
2. Build test
3. Update readme with description & example

PR #237, opened 2023-01-31
Feat/225 code editor remove deprecated

gitdallas:
#225 

also updated replace prop generator so it properly shows usage if called without args

### DIFF
--- a/packages/eslint-plugin-pf-codemods/test/rules/v5/codeEditor-remove-props.js
+++ b/packages/eslint-plugin-pf-codemods/test/rules/v5/codeEditor-remove-props.js
@@ -1,0 +1,26 @@
+const ruleTester = require('../../ruletester');
+const rule = require('../../../lib/rules/v5/codeEditor-remove-props');
+
+ruleTester.run("codeEditor-remove-props", rule, {
+  valid: [
+    {
+      code: `import { CodeEditor } from '@patternfly/react-core'; <CodeEditor no />`,
+    },
+    {
+      // No @patternfly/react-core import
+      code: `<CodeEditor entryDelay exitDelay maxWidth position toolTipText />`,
+    }
+  ],
+  invalid: [
+    {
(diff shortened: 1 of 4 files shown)


packages/pf-codemods/README.md
```diff
@@ -172,6 +172,10 @@ Out:
 <ClipboardCopyButton position="bottom" />
 ```
 
+### codeeditor-remove-deprecated-props [(#8624)](https://github.com/patternfly/patternfly-react/pull/8624)
+
+We've removed the deprecated `entryDelay`, `exitDelay`, `maxWidth`, `position`, and `toolTipText` props. This rule will remove them from your code and suggest that you use the tooltipProps prop instead.
```
thatblindgeye, line 177: Just missing an examples section for in/output.

thatblindgeye (changes requested):

thatblindgeye (approved):

packages/eslint-plugin-pf-codemods/test/rules/v5/codeEditor-remove-props.js
```diff
@@ -0,0 +1,26 @@
+const ruleTester = require('../../ruletester');
+const rule = require('../../../lib/rules/v5/codeEditor-remove-props');
+
+ruleTester.run("codeEditor-remove-props", rule, {
+  valid: [
+    {
+      code: `import { CodeEditor } from '@patternfly/react-core'; <CodeEditor no />`,
```
wise-king-sullyman, line 7: Not a blocker as I don't think there's any issues with it, but I'm curious what the reason is for the `no` here?

wise-king-sullyman (approved):
Looks good, one non-blocking question

packages/eslint-plugin-pf-codemods/test/rules/v5/codeEditor-remove-props.js
gitdallas: haha honestly idk, i may have used your generator and used "no" as the replacement value and forgot to remove here

Issue #242, opened 2023-02-07
Masthead, Pageheader - component type default changed


kmcfaul:
`component` in Masthead and `logoComponent` in Pageheader have been updated to change dynamically. The codemod should inform the user of the change and possibly also add the prop manually as an anchor to prevent DOM changes.

PR #243, opened 2023-02-07
chore(Masthead, PageHeader): component type default changed

kmcfaul:
Closes: #242

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/pageheader-update-logoComponent.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/pageheader-update-logoComponent.js
@@ -16,7 +16,7 @@ module.exports = {
               node,
               message: `The default PageHeader logoComponent type has be updated to switch between an anchor, button, and span based on whether a href, onClick or neither are present. It is also recommended to use Masthead in place of PageHeader.`,
               fix(fixer) {
-                return fixer.insertTextAfter(node.name, ` component="a"`);
+                return fixer.insertTextAfter(node.name, ` logoComponent="a"`);
               }
             });
           }
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/pageheader-update-logoComponent.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/pageheader-update-logoComponent.js
@@ -1,0 +1,27 @@
+const { getPackageImports } = require('../../helpers');
+
+// https://github.com/patternfly/patternfly-react/pull/8655
+module.exports = {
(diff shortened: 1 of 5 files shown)


packages/eslint-plugin-pf-codemods/lib/rules/v5/pageheader-update-logoComponent.js
```diff
@@ -0,0 +1,27 @@
...
+    return pageHeaderImport.length === 0 ? {} : {
+      JSXOpeningElement(node) {
+        if (pageHeaderImport.map(imp => imp.local.name).includes(node.name.name)) {
+          const componentAttr = node.attributes.find(n => n.name && n.name.name === 'logoComponent');
+          if (!componentAttr) {
+            context.report({
+              node,
+              message: `The default PageHeader logoComponent type has be updated to switch between an anchor, button, and span based on whether a href, onClick or neither are present. It is also recommended to use Masthead in place of PageHeader.`,
+              fix(fixer) {
+                return fixer.insertTextAfter(node.name, ` component="a"`);
```
gitdallas, line 19: i think this is a copy/paste typo and should be logoComponent.

gitdallas (changes requested):
~~Aside from the typo below, do we really want to put a fixer in here?~~

~~If we are defaulting to switch based on other logic, why do we want to add `="a"`? I'm trying to understand but perhaps most the time we'd rather they didn't specify the value? Also, this might cause issues if they were previously using a different value for `component`/`logoComponent`?~~

Maybe the above paragraph doesn't matter as I see you are checking for whether the property exists, and then using the previous default.. and then this would serve as a note to let them change if they want.

packages/eslint-plugin-pf-codemods/test/rules/v5/pageheader-update-logoComponent.js
```diff
@@ -0,0 +1,24 @@
...
+    },
+    {
+      // No @patternfly/react-core import
+      code: `<PageHeader />`,
+    }
+  ],
+  invalid: [
+    {
+      code:   `import { PageHeader } from '@patternfly/react-core'; <PageHeader />`,
+      output: `import { PageHeader } from '@patternfly/react-core'; <PageHeader component="a" />`,
```
gitdallas, line 17:
```suggestion
      output: `import { PageHeader } from '@patternfly/react-core'; <PageHeader logoComponent="a" />`,
```

gitdallas (changes requested):

gitdallas (approved):

thatblindgeye (approved):

PR #252, opened 2023-02-13
chore(Menu): moved arialabel from Menu to MenuList

### LINKED ISSUE #249: Menu/MenuList - moved aria-label from Menu to MenuList
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/8649

Codemod should either just warn consumers that the aria-label prop was removed from Menu and should be passed into MenuList instead, or we can try offering a fix to do that for them.

_Required actions:_
1. Build codemod
2. Build test

thatblindgeye:
Closes #249

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/menu-moved-ariaLabel.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/menu-moved-ariaLabel.js
@@ -1,0 +1,64 @@
+const { getPackageImports } = require("../../helpers");
+
+// https://github.com/patternfly/patternfly-react/pull/8649
+module.exports = {
+  meta: { fixable: "code" },
+  create: function (context) {
+    const menuImports = getPackageImports(
+      context,
+      "@patternfly/react-core"
+    ).filter((specifier) => specifier.imported.name == "Menu");
+    const getAriaLabel = (node) =>
+      node.openingElement.attributes.find(
+        (attr) => attr?.name?.name === "aria-label"
+      );
+
+    return menuImports.length === 0
+      ? {}
+      : {
+          JSXElement(node) {
+            const menuAriaLabel = getAriaLabel(node);
+
+            if (
+              menuImports
+                .map((imp) => imp.local.name)
+                .includes(node.openingElement.name.name) &&
+              menuAriaLabel
+            ) {
+              const menuLists = node.children.filter(
+                (child) => child?.openingElement?.name?.name === "MenuList"
+              );
+
+              context.report({
+                node,
+                message: `The aria-label prop has been removed from Menu and should be passed into MenuList instead. If using MenuGroup with the "label" prop, an aria-label on MenuList is not necessary.`,
+                fix(fixer) {
+                  const fixes = [];
+                  fixes.push(fixer.replaceText(menuAriaLabel, ""));
+
+                  if (
+                    menuLists.length === 1 &&
+                    !getAriaLabel(menuLists[0]) &&
+                    getPackageImports(context, "@patternfly/react-core").filter(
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin-pf-codemods/lib/rules/v5/menu-moved-ariaLabel.js
```diff
@@ -0,0 +1,64 @@
...
+                    const { range, selfClosing } = menuLists[0].openingElement;
+                    const rangeToReplace = selfClosing
+                      ? range[1] - 2
+                      : range[1] - 1;
+                    fixes.push(
+                      fixer.replaceTextRange(
+                        [rangeToReplace, rangeToReplace],
+                        ` aria-label="${menuAriaLabel.value.value}"`
+                      )
+                    );
```
gitdallas, line 55: to avoid leaving an extra space with self closing tags you could create a range of 1 to trim that out or remove the added space in the replacement.. although i think i'd prefer to use insertTextAfter the name node, something like..

```suggestion
                    fixes.push(
                      fixer.insertTextAfter(menuLists[0].openingElement.name, ` aria-label="${menuAriaLabel.value.value}"`);
                    );
```

packages/eslint-plugin-pf-codemods/lib/rules/v5/menu-moved-ariaLabel.js
```diff
@@ -0,0 +1,64 @@
...
+
+              context.report({
+                node,
+                message: `The aria-label prop has been removed from Menu and should be passed into MenuList instead. If using MenuGroup with the "label" prop, an aria-label on MenuList is not necessary.`,
+                fix(fixer) {
+                  const fixes = [];
+                  fixes.push(fixer.replaceText(menuAriaLabel, ""));
+
+                  if (
+                    menuLists.length === 1 &&
```
gitdallas, line 40: we wouldn't want to do anything with multiple lists? i'm not really familiar with how these might be set up

packages/eslint-plugin-pf-codemods/test/rules/v5/menu-moved-ariaLabel.js
```diff
@@ -0,0 +1,84 @@
...
+      errors: [
+        {
+          message: `The aria-label prop has been removed from Menu and should be passed into MenuList instead. If using MenuGroup with the "label" prop, an aria-label on MenuList is not necessary.`,
+          type: "JSXElement",
+        },
+      ],
+    },
+    {
+      code: `import { Menu, MenuList } from '@patternfly/react-core'; <Menu aria-label="tester"><MenuList /></Menu>`,
+      output: `import { Menu, MenuList } from '@patternfly/react-core'; <Menu ><MenuList  aria-label="tester"/></Menu>`,
```
gitdallas, line 34: with the rule change suggestion..

```suggestion
      output: `import { Menu, MenuList } from '@patternfly/react-core'; <Menu ><MenuList aria-label="tester"/></Menu>`,
```

wonder if we should also add a non-self closing MenuList as you addressed that in the rules

gitdallas (changes requested):
lgtm.. just a suggestion and a couple questions

packages/eslint-plugin-pf-codemods/lib/rules/v5/menu-moved-ariaLabel.js
thatblindgeye: My thought was if there are multiple lists within a single Menu, consumers should go through them to apply an accurate aria-label based on the context rather than each list being given the same label.

gitdallas (approved):

wise-king-sullyman (approved):

PR #256, opened 2023-02-13
chore(Table): collapseAllAriaLabel type update warning

### LINKED ISSUE #255: Table - warn that ThExpandType collapseAllAriaLabel type updated
Codemod for https://github.com/patternfly/patternfly-react/pull/8634

Though the type updated, no fix should be required. So just needs a warning to let people know if they were working around the `''` type by casting.

kmcfaul:
Closes #255

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/table-warn-thExpandType.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/table-warn-thExpandType.js
@@ -1,0 +1,24 @@
+const { getPackageImports } = require("../../helpers");
+
+// https://github.com/patternfly/patternfly-react/pull/8634
+module.exports = {
+  meta: {},
+  create: function (context) {
+    const tableImport =  getPackageImports(context, '@patternfly/react-core')
+    .filter(specifier => specifier.imported.name == 'Table' || specifier.imported.name === 'Th');
+
+    return tableImport.length === 0 ? {} : {
+      JSXOpeningElement(node) {
+        if (tableImport.map(imp => imp.local.name).includes(node.name.name)) {
+          const hasCollapseAllAriaLabel = node.attributes.find(n => n.name && n.name.name === 'collapseAllAriaLabel');;
+          if (hasCollapseAllAriaLabel) {
+            context.report({
+              node,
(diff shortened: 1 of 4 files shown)


packages/pf-codemods/README.md
```diff
@@ -542,6 +542,27 @@ Out:
 <Spinner  />
 ```
 
+### table-warn-thExpandType [(#8634)](https://github.com/patternfly/patternfly-react/pull/8634)
+
+`collapseAllAriaLabel` on `ThExpandType` has been updated to a `string` from `''`. Workarounds casting this property to an empty string are no longer required.
+
+#### Examples
```
gitdallas, line 549: I don't think that we need to add examples when there are no fixers.

packages/eslint-plugin-pf-codemods/lib/rules/v5/table-warn-thExpandType.js
```diff
@@ -0,0 +1,24 @@
...
+  meta: {},
+  create: function (context) {
+    const tableImport =  getPackageImports(context, '@patternfly/react-table')
+    .filter(specifier => specifier.imported.name == 'Table' || specifier.imported.name === 'Th');
+
+    return tableImport.length === 0 ? {} : {
+      JSXOpeningElement(node) {
+        if (tableImport.map(imp => imp.local.name).includes(node.name.name)) {
+          const hasCollapseAllAriaLabel = node.attributes.find(n => n.name && n.name.name === 'collapseAllAriaLabel');;
+          if (hasCollapseAllAriaLabel) {
```
gitdallas, line 14: as we are getting the attribute rather than a boolean, probably name it as such, remove extra `;`

```suggestion
          const collapseAllAriaLabel = node.attributes.find(n => n.name && n.name.name === 'collapseAllAriaLabel');
          if (collapseAllAriaLabel) {
```
bonus might be if we check if it's actually an `as` expression for the value. `collapseAllAriaLabel.?value.?expression?.type` and only apply the warning if it is an `TSAsExpression`

this way the warning wouldn't keep popping up after they fix it and recheck.

gitdallas (changes requested):
couple suggestions

gitdallas (approved):

PR #260, opened 2023-02-13
feat(OverflowMenuDropdownItem): updated to use Next DropdownItem

### LINKED ISSUE #258: feat(OverflowMenu): Use next dropdownItems
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/8359

Codemod should warn that OverflowMenuDropdownItem uses the Next DropdownItem under the hood, and that any use of non-Next Dropdown components should be updated to use Next Dropdown components (might be too much of a wildcard to offer a fix for this). Additionally, we can offer a fix to update the `index` prop for OverflowMenuDropdownItem to `itemId`.

_Required actions:_
1. Build codemod
2. Build test

thatblindgeye:
Closes #258 

Opted to create separate codemods since one seemed more of a "warning" than an error, but originally I did have them combined as a single codemod with a dynamic message if we want to do that instead.

Also fixed an Application Launcher codemod as it wouldn't have ran if ApplicationLauncher was anything but the first import.


gitdallas (approved):

packages/pf-codemods/README.md
```diff
@@ -791,6 +791,30 @@ function toggleDropdown(_event, isOpen) {};
...
+```jsx
+<OverflowMenuDropdownItem index={0}>
+```
+
+Out:
+
+```jsx
+<OverflowMenuDropdownItem itemId={0}>
+```
+
```
gitdallas, line 817: I don't think this is necessary

gitdallas (commented):
Think we can get rid of examples if in/out are the same, otherwise lgtm

packages/eslint-plugin-pf-codemods/lib/rules/v5/overflowMenuDropdownItem-renamed-prop.js
```diff
@@ -0,0 +1,41 @@
+const { getPackageImports } = require("../../helpers");
+
+// https://github.com/patternfly/patternfly-react/pull/8359
+module.exports = {
+  meta: { fixable: "code" },
+  create: function (context) {
```
wise-king-sullyman, line 6: Not a blocker, but I'm curious why you didn't just use the rename props helper?

wise-king-sullyman (commented):
Agree with Dallas, and one question:

packages/eslint-plugin-pf-codemods/lib/rules/v5/overflowMenuDropdownItem-renamed-prop.js
gitdallas: may have been because of the custom message - but one of the helpers allows for that as well.

gitdallas (approved):

packages/pf-codemods/README.md
gitdallas: oops, i'm blind

wise-king-sullyman: Oh man, yeah also totally missed that 😆

PR #263, opened 2023-02-14
chore(misc): warn about updated onToggle api

### LINKED ISSUE #245: Misc - update dropdown/select/applauncher toggle props
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/issues/8664

kmcfaul:
Closes #245


packages/eslint-plugin-pf-codemods/test/rules/v5/warn-event-api.js
```diff
@@ -0,0 +1,86 @@
+const ruleTester = require('../../ruletester');
+const rule = require('../../../lib/rules/v5/warn-event-api');
+
+ruleTester.run("warn-event-api", rule, {
+  valid: [
```
gitdallas, line 5: not really necessary as the rule is only on import declarations.. could test `import { DropdownToggle } from '@foo/bar'` but no biggie

gitdallas (approved):

packages/eslint-plugin-pf-codemods/test/rules/v5/warn-event-api.js
gitdallas: another option would be to put the warnings on the jsx elements like here https://github.com/patternfly/pf-codemods/pull/267

this would point them in the code closer to where they'd want to make changes, but again, not a blocker

wise-king-sullyman (approved):
Also not a blocker, but as @thatblindgeye [mentioned on my Nav warning PR](https://github.com/patternfly/pf-codemods/pull/266#discussion_r1106212156) we probably want to standardize on not having "warn" as the first word in the rule name.

If you don't think you have time to address today though just lmk and we can worry about that in a followup and I'll merge this PR.

kmcfaul:
renamed to `onToggle-warn-event`, since it covers multiple components.

gitdallas:
After coming across this in cost... I'm wondering if we can look at attributes for the actual onToggle?

And possibly even a fixer... could convert

`onToggle={this.handleOnToggle}`
to
`onToggle={(event, isOpen) => this.handleOnToggle(isOpen)}`

and

`onToggle={ (isOpen) => { ... } }`
to
`onToggle={ (event, isOpen) => { ... }}`

thoughts @kmcfaul @wise-king-sullyman ?

wise-king-sullyman:
That definitely sounds like it would be nice. Would have to dig into it to see how feasible it would be.

PR #265, opened 2023-02-14
fix(Charts): fixed broken tooltip examples

### LINKED ISSUE #234: Charts - broken legend Tooltip
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/8592

The PR above fixes a broken chart example due to a `Tooltip` update. This particular example demonstrates an approach for applying tooltips to legend text using a custom label component.

The `Tooltip` popper fails to render because it now outputs an HTML `div` tag in the chart's SVG. In order for HTML tags to render properly, they must be wrapped by `foreignObject`.

Screen Shot 2023-01-27 at 5 32 31 PM

Note that the `x` and `y` coordinates are custom properties and may not apply to all instances. We may only be able to output a warning and/or document this change in the README?

Required actions:
1. Build codemod
2. Build test

thatblindgeye:
Closes #234 

Also updated some logic for the getResizeObserver file so that if there are two separate import statements from react-charts, only getResizeObserver will be updated to react-core (before it would update all imports from react-charts if getResizeObserver was found).

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/charts-resizeObserver-import.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/charts-resizeObserver-import.js
@@ -13,7 +13,13 @@ module.exports = {
       ? {}
       : {
           ImportDeclaration(node) {
-            if (node.source.value === "@patternfly/react-charts") {
+            if (
+              node.specifiers.filter((specifier) =>
+                chartResizeImport
+                  .map((imp) => imp.local.name)
+                  .includes(specifier.imported.name)
+              ).length
+            ) {
               context.report({
                 node,
                 message: `The getResizeObserver function has been removed from react-charts and should be imported from react-core instead.`,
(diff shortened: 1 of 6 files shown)


packages/eslint-plugin-pf-codemods/lib/rules/v5/charts-resizeObserver-import.js
```diff
@@ -13,7 +13,13 @@ module.exports = {
       ? {}
       : {
           ImportDeclaration(node) {
-            if (node.source.value === "@patternfly/react-charts") {
+            if (
```
gitdallas, line 16: good catch

gitdallas (commented):
would be cool to have this actually check that we have a charts component and not already using `foreignObject`, perhaps a future enhancement if we have some time

gitdallas (approved):

dlabrecq (approved):

PR #266, opened 2023-02-14
chore(Nav): added warning for flyout dom position changes

### LINKED ISSUE #262: Nav - Flyouts now append inline
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/8628

The placement of flyouts in the DOM has changed, we probably want to warn that snapshots/selectors may need updating.

_Required actions:_
1. Build codemod
2. Build test

wise-king-sullyman:
Closes #262


packages/eslint-plugin-pf-codemods/index.js
```diff
@@ -21,6 +21,7 @@ const warningRules = [
   "react-dropzone-warn-upgrade",
   "table-warn-thExpandType",
   "tabs-warn-children-type-changed",
+  "warn-nav-flyouts-now-inline",
```
thatblindgeye, line 24: Nit: just to align with other rule names (the `warn-key-codes-removed` may also be able to be updated)

```suggestion
  "nav-warn-flyouts-now-inline",
```

wise-king-sullyman: Addressed.

Also addressed warn-key-codes-removed, which ended up raising an error rather than a warning anyways so I adjusted the rule name and description.

thatblindgeye (approved):

packages/eslint-plugin-pf-codemods/lib/rules/v5/nav-warn-flyouts-now-inline.js
```diff
@@ -0,0 +1,33 @@
...
+
+// https://github.com/patternfly/patternfly-react/pull/8628
+module.exports = {
+  meta: {},
+  create: function (context) {
+    const imports = getPackageImports(context, "@patternfly/react-core").filter(
+      (specifier) => specifier.imported.name === "Nav"
+    );
+
+    return imports === 0
```
thatblindgeye, line 11: Just noticed, but this should be:

```suggestion
    return imports.length === 0
```

kmcfaul (approved):

wise-king-sullyman:
@kmcfaul as I was resolving conflicts I realized that we missed changing the rule name in the readme of your last PR, I went ahead and updated it. LMK if that doesn't look right for some reason.

thatblindgeye (approved):

packages/eslint-plugin-pf-codemods/lib/rules/v5/nav-warn-flyouts-now-inline.js
```diff
@@ -0,0 +1,33 @@
...
+    const imports = getPackageImports(context, "@patternfly/react-core").filter(
+      (specifier) => specifier.imported.name === "Nav"
+    );
+
+    return imports.length === 0
+      ? {}
+      : {
+          JSXOpeningElement(node) {
+            const hasNav = imports
+              .map((imp) => imp.local.name)
```
gitdallas, line 16: as we should only have the `Nav` import, i think we should just call the `import` `navImport` with a `find` instead of `filter` and not have to map on this

gitdallas (approved):
lgtm.. small nit not really blocking

gitdallas (approved):

PR #280, opened 2023-02-16
fix(index): dynamically resolve path to rules

### LINKED ISSUE #279: Rules not being properly loaded
It appears that the eslint-plugin-pf-codemods rules aren't properly loading when pf-codemods is actually used.

wise-king-sullyman:
Closes #279


gitdallas (approved):
ty

nicolethoen (approved):

PR #284, opened 2023-02-17
fix issue with spread attributes breaking codemod execution (not a codemod)

gitdallas:


packages/eslint-plugin-pf-codemods/lib/rules/v5/toggle-remove-isPrimary.js
```diff
@@ -10,7 +10,7 @@ const { getPackageImports } = require('../../helpers');
       return toggleImport.length === 0 ? {} : {
         JSXOpeningElement(node) {
           if (toggleImport.map(imp => imp.local.name).includes(node.name.name)) {
-            const isPrimaryAttr = node.attributes.find(n => n.name && n.name.name === 'isPrimary');;
+            const isPrimaryAttr = node.attributes.find(a => a.name?.name === 'isPrimary');;
```
jeffpuzzo, line 13: nit: Seems like there's a lot of inconsistency in what `a` is called here. Sometimes `a`, `node`, `nodeAttr`, `attribute`. I think it would best to stick with 1, and something more verbose like `nodeAttr` or `attribute` would be my preference.

jeffpuzzo (approved):

packages/eslint-plugin-pf-codemods/lib/rules/v5/toggle-remove-isPrimary.js
gitdallas: with short functions i tend to stick with short names (based on reading "clean code" many years ago lol), i like the `a` here as to me it contrasts it against the `node` it is part of.

Issue #288, opened 2023-02-21
Get codemods to work with alias imports


gitdallas:
Our codemods will currently not work if a consumer does something like

`import { Popover as PfPopover } from ‘@patternfly/react-core’;`

need to find the fix and make sure existing codemods work.

PR #290, opened 2023-02-21
fix(various): several codemods were failing if consumers used alias i…

gitdallas:
#288

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/helpers.js
+++ b/packages/eslint-plugin-pf-codemods/lib/helpers.js
@@ -8,7 +8,9 @@ function getPackageImports(context, packageName) {
 
 function renameProps0(context, imports, node, renames) {
   if (imports.map(imp => imp.local.name).includes(node.name.name)) {
-    const renamedProps = renames[node.name.name];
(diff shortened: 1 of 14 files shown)


thatblindgeye (approved):
This looks good. One nit is the inconsistency between what the output message is in the test, i.e. `popover-removeProps` uses the full Popover in the message, but `remove-removeFindDomNode` uses the alias "Pop". Not a huge thing, but might be worth fixing (wouldn't block this PR over it, though).

wise-king-sullyman (approved):
LGTM

gitdallas:
> This looks good. One nit is the inconsistency between what the output message is in the test, i.e. `popover-removeProps` uses the full Popover in the message, but `remove-removeFindDomNode` uses the alias "Pop". Not a huge thing, but might be worth fixing (wouldn't block this PR over it, though).

I saw this as well. The cases with that use the `renameProps` helper are using the local (ei: "Pop") while `popover-remove-props` is using "Popover" hardcoded in the message string. I wonder which is best? "Popover" is the actual component, but when we point to a line in the code the user would see the local name. A bit of an edge case though since alias are probably not used too often on these. Also a consumer should be able to identify either way.

PR #306, opened 2023-03-06
feat(DataList): added event param to callback

### LINKED ISSUE #289: DataList - added event param to callback
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/8723

Warn that an event param has been added to the onSelectDataListItem prop

_Required actions:_
1. Build codemod
2. Build test

### LINKED ISSUE #295: Misc - toggle props codemod enhancement
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/issues/8664

Enhance [original issue](https://github.com/patternfly/pf-codemods/issues/245) so that the codemods only trigger if there is an applicable prop (`onToggle`), also we could look into a fixer:

`onToggle={this.handleOnToggle}`
to
`onToggle={(event, isOpen) => this.handleOnToggle(isOpen)}` 
(although we'd have to make sure that `handleOnToggle` was actually taking the parameter, can we do this?)

and

`onToggle={ (isOpen) => { ... } }`
to
`onToggle={ (event, isOpen) => { ... }}`

other places it might not need a fix at all, if inline function and no params given, can ignore:

        isOpen={isOpen}
        onToggle={() => setIsOpen(!isOpen)}

_Required actions:_
1. Build codemod
2. Build test

thatblindgeye:
Closes #289, closes #295

The update made to SelectableRowObject in the linked PR for this codemod shouldn't be needed here, as https://github.com/patternfly/patternfly-react/issues/8423 will require a codemod that will handle that.

Also made an update to the onToggle codemod similar to this.


gitdallas (approved):

thatblindgeye:
@gitdallas for now I ust copy+pasted the logic from the onToggle codemod to the DataList one. I can do something similar to my other 2 open PRs. Figure we can create a possible helper as a followup.

packages/eslint-plugin-pf-codemods/lib/rules/v5/onToggle-updated-paramaters.js
```diff
@@ -0,0 +1,87 @@
...
+            const { type, params, name } =
+              onToggleProp?.value?.expression || {};
+
+            if (
+              imports.map((imp) => imp.local.name).includes(node.name.name) &&
+              ((type === "ArrowFunctionExpression" && params?.length === 1) ||
+                type === "Identifier")
+            ) {
+              context.report({
+                node,
```
gitdallas, line 37: think we could strip this out to be reused? seems like it will be in several places already

packages/eslint-plugin-pf-codemods/lib/rules/v5/dataList-updated-callback.js
```diff
@@ -0,0 +1,78 @@
...
+            const onSelectDataListItemProp = node.attributes.find(
+              (attr) => attr.name.name === "onSelectDataListItem"
+            );
+
+            const { type, params, name } =
+              onSelectDataListItemProp?.value?.expression || {};
+
+            if (
+              imports.map((imp) => imp.local.name).includes(node.name.name) &&
+              ((type === "ArrowFunctionExpression" && params?.length === 1) ||
```
gitdallas, line 24: consider class level functions too for reporting, regardless of if we can apply a fixer

packages/eslint-plugin-pf-codemods/test/rules/v5/onToggle-updated-paramaters.js
```diff
@@ -0,0 +1,88 @@
...
+      output: `import { ${component} } from '@patternfly/react-core'; const onToggle = (_event, isOpen) => {}; <${component} onToggle={onToggle} />;`,
+      errors: [
+        {
+          message: `${component} onToggle prop has been updated to include the event parameter as its first parameter. onToggle handlers may require an update.`,
+          type: "JSXOpeningElement",
+        },
+      ],
+    })),
+    ...onToggleAPIUpdateList.map((component) => ({
+      code: `import { ${component} } from '@patternfly/react-core'; function onToggle() {}; <${component} onToggle={onToggle} />;`,
```
gitdallas, line 68: would we be able to ignore this case?

gitdallas (changes requested):
overall this is awesome, a few comments/nits

packages/eslint-plugin-pf-codemods/lib/rules/v5/onToggle-updated-paramaters.js
gitdallas: i see you left a comment regarding this, i'll leave that up to you

wise-king-sullyman (approved):
This is like magic

thatblindgeye:
@gitdallas @wise-king-sullyman refactored the logic in both codemods as a (currently very basic, specific) helper function. Should be a lot easier for similar codemods for adding that "event" param (or any param really, as long as its meant to be the first one).

Figure this could be a building block for more flexible things down the line

wise-king-sullyman (approved):
🔥 🔥 🔥 🔥 🔥

PR #308, opened 2023-03-06
feat(Drawer): added event param to onResize

### LINKED ISSUE #292: Drawer - added event param to onResize callback
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/8736

warn that an event param has been added to the onResize callback

_Required actions:_
1. Build codemod
2. Build test

thatblindgeye:
Closes #292


gitdallas:
should use the new helper function as well, right?

thatblindgeye:
Yeah I'd need to rebase once my DataListCheck goes in as that should resolve some failing tests locally. I bumped this to the next milestone, though since I'll be out the rest of the week.

packages/pf-codemods/README.md
```diff
@@ -412,6 +412,12 @@ Out:
 <Divider orientation={{ default: "vertical" }} />
 ```
 
+### drawer-warn-updated-callback [(#8736)](https://github.com/patternfly/patternfly-react/pull/8736)
+
+We've updated the `onResize` prop for Drawer to include the `event` as its first parameter. Handlers may require an update.
+
+This rule will raise a warning, but will not make any changes.
```
gitdallas, line 419: 😏

gitdallas (changes requested):
readme could be updated after the fixers

packages/pf-codemods/README.md
thatblindgeye: Good catch. Also updated the file names to be more specific (Drawer > DrawerPanelContent)

gitdallas (approved):

wise-king-sullyman (approved):

PR #316, opened 2023-03-06
chore(DataList): remove itemOrder prop

### LINKED ISSUE #303: DataList - remove itemOrder prop
**This issue was raised when testing pf5 alphas in Keycloak. It was accounted for in build errors but not in codemods.**

Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/8388

Remove usage of deprecated props:

itemOrder

Required actions:

- [x] Build codemod
- [x] Build test
- [x] Update readme with description & example

jenny-s51:
Closes #303


gitdallas (approved):

packages/eslint-plugin-pf-codemods/lib/rules/v5/datalist-remove-itemOrder.js
```diff
@@ -0,0 +1,11 @@
+const { renameProp } = require('../../helpers');
+
+https://github.com/patternfly/patternfly-react/pull/8388/files
```
thatblindgeye, line 3: Just need to comment this out

thatblindgeye (changes requested):
One quick change below, but was also wondering whether it'd make sense to combine this with the existing `datalist-remove-ondrags` codemod, and just have a single `datalist-remove-props` file. Would just need to have a dynamic error message for when an `onDragX` prop is passed to include the "Implement drag and drop using the DragDrop component instead" verbiage.

This isn't a blocker, though, and could maybe be a followup if we did want to do that.

packages/pf-codemods/README.md
```diff
@@ -311,22 +311,22 @@ Out:
 <ClipboardCopy  />
 ```
 
-### datalist-remove-ondrags [(#8388)](https://github.com/patternfly/patternfly-react/pull/8388)
+### datalist-remove-props [(#8388)](https://github.com/patternfly/patternfly-react/pull/8388)
 
-We've removed the deprecated `onDragFinish`, `onDragStart`, `onDragMove`, and `onDragCancel` props. This rule will remove them and suggest the user use the DragDrop component.
+We've removed the deprecated `onDragFinish`, `onDragStart`, `onDragMove`, and `onDragCancel`, and `itemOrder` props. This rule will remove them and suggest the user use the DragDrop component.
```
thatblindgeye, line 316: Not a blocker, but just to separate the onDrag from itemOrder:

```suggestion
We've removed the deprecated `onDragFinish`, `onDragStart`, `onDragMove`, `onDragCancel`, and `itemOrder` props from DataList.

In addition to removing these props, this rule will suggest using the DragDrop component instead of the `onDrag...` props.
```

Or something along those lines.

thatblindgeye (commented):
🔥 Looks great! Had one other comment below, but it's not a blocker so let me know what you think

thatblindgeye (approved):

gitdallas (approved):

PR #317, opened 2023-03-06
feat(Label): remove isTruncated from Label

### LINKED ISSUE #273: Label - Remove isTruncated prop
Follow up to breaking change issue https://github.com/patternfly/patternfly-react/issues/8258

The label will truncate by default.  Remove the `isTruncated` prop.

_Required actions:_
1. Build codemod
2. Build test

gitdallas:
Closes #273

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/label-remove-isTruncated.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/label-remove-isTruncated.js
@@ -1,0 +1,11 @@
+const { renameProp } = require('../../helpers');
+
+  // https://github.com/patternfly/patternfly-react/pull/8771
+  module.exports = {
+    meta: { fixable: 'code' },
+    create: renameProp(
+      'Label',
+      {'isTruncated': ''},
+      node =>  `isTruncated prop has been removed for ${node.name.name}. This is now the default behavior. Note that there is also a new property (maxTextWidth) to customize when truncation will occur.`
+    ),
+  };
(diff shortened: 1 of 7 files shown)


gitdallas:
@nicolethoen  - I wonder if we should worry about putting a "warning" for cases where consumers did NOT have isTruncated, or isTruncated=false... as with it being the default, their DOM will look slightly different.

packages/eslint-plugin-pf-codemods/lib/rules/v5/label-remove-isTruncated.js
```diff
@@ -0,0 +1,11 @@
+const { renameProp } = require('../../helpers');
+
+  // https://github.com/patternfly/patternfly-react/pull/8771
+  module.exports = {
+    meta: { fixable: 'code' },
+    create: renameProp(
+      'Label',
+      {'isTruncated': ''},
+      node =>  `isTruncated prop has been removed for ${node.name.name}. This is now the default behavior.`
```
thatblindgeye, line 9: Could also mention the `textMaxWidth` prop as a means to customize when truncation occurs.

thatblindgeye (commented):
Looks good. In regards to comment above, I'd agree that it may be worth providing a message for when `isTruncated` isn't passed. Also had a comment below, let me know what you think

packages/eslint-plugin-pf-codemods/lib/rules/v5/label-remove-isTruncated.js
gitdallas: i thought about this, and included it in the readme. I'm not sure it's worth putting here as if they currently have it implemented with `isTruncated` it's going to be working as they want in that implementation (maybe a wrapper of some sort). I don't know that I would expect products to refactor at this point.

But... I suppose it wouldn't hurt to mention.

thatblindgeye: Yeah I don't feel super strong about it, figured I'd mention it in case, though.

thatblindgeye (approved):

wise-king-sullyman (approved):

PR #314, opened 2023-03-06
chore(Tooltip): renamed reference prop to triggerRef

adamviktora:
Closes #304, relates to https://github.com/patternfly/patternfly-react/pull/8733

PR #318, opened 2023-03-06
chore(Popover, Tooltip): renamed reference prop to triggerRef

adamviktora:
Closes https://github.com/patternfly/pf-codemods/issues/296 and #304, relates to https://github.com/patternfly/patternfly-react/pull/8733

(merging two issues into this PR -> PR #314 was closed)


gitdallas:
according to comments on #314  perhaps we should combine these

gitdallas (changes requested):
we can actually include this into a single rule file, similar to `toolbarVarious-remove-alignment` or `remove-removeFindDomNode`, then maybe just call it `various-rename-reference` for example.

jenny-s51 (approved):
Looks and runs great! Nice work @adamviktora 👍

gitdallas (approved):

thatblindgeye (approved):

PR #320, opened 2023-03-07
chore(Pagination): rename ToggleTemplateProps

### LINKED ISSUE #305: Pagination - ToggleTemplateProps: rename not handled in import
Follow up to https://github.com/patternfly/pf-codemods/pull/201/files

Rule should be updated to handle prop renaming via import

_Required actions:_
1. Update codemod
2. Update test

kmcfaul:
Closes #305

* Checks imports to rename `ToggleTemplateProps` to `PaginationToggleTemplateProps`.
* Checks `toggleTemplate` prop and if it is passed a function, and that function has a type annotation, the annotation is also updated to `PaginationToggleTemplateProps`.

Put this in a separate rule tester than the current `pagination-remove-ToggleTemplateProps`. We might be able to remove the other rule but I left it in for now as it checks direct usage.


gitdallas:
Actually it looks like the previous PR (my own) was just wrong. Looks like I thought we were renaming a property rather than a type, and that property doesn't even exist. There is no `ToggleTemplateProps` or `PaginationToggleTemplateProps` on `Pagination`, only `toggleTemplate` which takes that type.

The old rule could be removed and the readme should be updated as well.

gitdallas (approved):
One thing I noticed is that with an alias import, this would basically remove the alias.

So `ToggleTemplateProps as TogProps ... : TogProps` would become `PaginationToggleTemplateProps ... : PaginationToggleTemplateProps`

I think this is pretty edge case-y though, and shouldn't break code so I'm not too concerned.

packages/pf-codemods/README.md
```diff
@@ -639,6 +621,26 @@ Out:
 />
 ```
 
+### pagination-rename-toggleTemplateProps [(#8134)](https://github.com/patternfly/patternfly-react/pull/8134)
+
+We've renamed the `ToggleTemplateProps` prop and to `PaginationToggleTemplateProps`.
```
jenny-s51, line 626: Following @gitdallas 's comment above regarding how this wasn't a prop and is instead used as a type!
```suggestion
We've renamed `ToggleTemplateProps` to `PaginationToggleTemplateProps`.
```

jenny-s51 (changes requested):
This is awesome @kmcfaul LGTM 🎉 Just needs conflicts resolved, and left a small suggestion regarding verbiage. Should we also add a test case to `test.tsx`? I was able to test successfully with the inputs you added to the README!

thatblindgeye (approved):
Just needs a rebase

jenny-s51 (approved):

PR #321, opened 2023-03-07
feat(ApplicationLauncher): added event params to onSearch and onFavorite

### LINKED ISSUE #313: Application launcher - warn that an event param has been added to callbacks
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/8756

Warn that the onFavorite and onSearch callbacks have had an event added as their first parameter.

_Required actions:_
1. Build codemod
2. Build test

wise-king-sullyman:
Closes #313


packages/eslint-plugin-pf-codemods/test/rules/v5/applicationLauncher-updated-params.js
```diff
@@ -0,0 +1,90 @@
+const ruleTester = require('../../ruletester');
+const rule = require('../../../lib/rules/v5/applicationLauncher-updated-params');
+const propNames = ['onFavorite', 'onSearch'];
+
+const ruleRunner = (prop) => {
```
gitdallas, line 5: could we just do `propNames.forEach(  (prop) => {` directly? so I don't have to wonder what's going on until i scroll to the bottom 😆

wise-king-sullyman: What's wrong with having some adventure every and then? 😂

wise-king-sullyman: That being said, given the temporary nature of codemods I wonder a bit about extracting these event param tests into a reusable helper function.

I'd typically be against that sort of thing since it decreases test readability, but that tradeoff would probably be worth doing for codemods since we won't be maintaining them long term and it would speed up development.

WDYT?

gitdallas: if it's set up right and we trust the helpers, it shouldn't reduce readability.. might make sense to have another helper in the tests dir, up to you if you wanna take a shot at that

gitdallas (approved):
re-request if you plan to create a test helper, but otherwise lgtm

wise-king-sullyman:
First swing at a helper added, not my proudest work.... but it gets the job done. WDYT @gitdallas ?

gitdallas:
@wise-king-sullyman - it's very specific, down to the "_event" being hard coded in.. but does the job for reducing a lot of duplicate code in 4-5 codemods, although I think in at least one case there is one that has more than one parameter before adding the new one that it wouldn't work for. i don't think we need to export the get..Tests functions.

wise-king-sullyman:
Yeah my idea was for it to be a counterpart specifically for the `addCallbackParam` helper Eric made, which also has the message hardcoded with _event and such. If you think this is too niche though I can see how I can open it up.

gitdallas:
maybe we'll deal with opening it up if we run into another case down the road. but should we use this where it can be now?

wise-king-sullyman:
Reverted the helper in this PR and created #334 for the test helper. Once either of these are merged I'll update the other as needed.

gitdallas (approved):

thatblindgeye (changes requested):
Could we add to the test.tsx file as well?

thatblindgeye (approved):

gitdallas (approved):

PR #338, opened 2023-03-14
feat(Banner): updated variant types

### LINKED ISSUE #331: Banner - API changes
Follow up to breaking change issue https://github.com/patternfly/patternfly-react/issues/8204

 updated variant names to reference colors instead of status names

_Required actions:_
1. Build codemod
2. Build test

thatblindgeye:
Closes #331 

Because of how much variety there may be when it comes to passing some variable/function reference to the `variant` prop, I opted to only include a fix for when a string is passed in. Otherwise an error message will be output if the `variant` prop is used at all (even if "default" is being passed for some reason, it lets consumers know the other variant types have changed)

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/banner-update-variant.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/banner-update-variant.js
@@ -24,14 +24,18 @@ module.exports = {
                 (attribute) => attribute.name.name === "variant"
               );
 
-              if (variantProp) {
+              if (
+                variantProp &&
+                (variantMap[variantProp.value?.value] ||
+                  variantProp.value?.type !== "Literal")
+              ) {
                 context.report({
                   node,
                   message: `The "variant" prop type for Banner has been updated. "default" is still a valid value, but the previous status values of "info", "success", "warning", and "danger" have been replaced with color values of "blue", "green", "gold", and "red", respectively.`,
                   fix(fixer) {
                     const fixes = [];
 
-                    if (variantMap[variantProp.value?.value]) {
+                    if (variantMap[variantProp.value.value]) {
                       fixes.push(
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin-pf-codemods/lib/rules/v5/banner-update-variant.js
```diff
@@ -0,0 +1,51 @@
...
+      success: "green",
+    };
+
+    return !bannerImport
+      ? {}
+      : {
+          JSXOpeningElement(node) {
+            if (bannerImport.local.name === node.name.name) {
+              const variantProp = node.attributes.find(
+                (attribute) => attribute.name.name === "variant"
```
gitdallas, line 24: can you also check that the value is in variantMap object keys?

packages/eslint-plugin-pf-codemods/test/rules/v5/banner-update-variant.js
```diff
@@ -0,0 +1,37 @@
...
+  { info: "blue" },
+  { danger: "red" },
+  { warning: "gold" },
+  { success: "green" },
+];
+
+ruleTester.run("banner-update-variant", rule, {
+  valid: [
+    {
+      code: `import { Banner } from '@patternfly/react-core'; <Banner />`,
```
gitdallas, line 14: should add `variant="blue"` for example and make that valid

gitdallas (changes requested):

thatblindgeye:
@gitdallas updated the logic a bit. Now an error should be caught only if:

- variant is passed a Literal ("danger", etc), and that value is anything but "default"
- variant is passed anything but a Literal (just to provide a warning in case anything more complicated is going on)

gitdallas (approved):

PR #340, opened 2023-03-16
chore(EmptyStateIcon): removed props, icon prop required

### LINKED ISSUE #336: EmptyStateIcon - remove props 'variant' and 'component'; 'icon' prop is now mandatory
Follow up to breaking change PR [[link]](https://github.com/patternfly/patternfly-react/pull/8737)

-- props 'variant' and 'component' were removed
-- 'icon' prop is now mandatory

_Required actions:_
1. Build codemod
2. Build test

adamviktora:
Closes #336


packages/eslint-plugin-pf-codemods/lib/rules/v5/emptyStateIcon-icon-required.js
```diff
@@ -0,0 +1,30 @@
...
+      ? {}
+      : {
+          JSXOpeningElement(node) {
+            if (imports.find((imp) => imp.local.name === node.name.name)) {
+              const iconPropIncluded = node.attributes
+                .filter((attr) => attr.name)
+                .map((attr) => attr.name.name)
+                .includes("icon");
+
+              if (!iconPropIncluded) {
```
gitdallas, line 20:
```suggestion
              if ( node.attributes.find( a => ['icon', 'component'].includes(a.name?.name))) {
```

I wonder if we should also check for component here like above, as the other rule will convert component into icon... 

as it stands now, if they were using component it would say "icon is required", "component is now icon" and also convert it to icon with a fixer..

what do you guys think @thatblindgeye @wise-king-sullyman ?

currently
`70|  <EmptyStateIcon component={Spinner} variant="container"/>`
will give
image

gitdallas (changes requested):

packages/eslint-plugin-pf-codemods/lib/rules/v5/emptyStateIcon-icon-required.js
```diff
@@ -0,0 +1,30 @@
+const { getPackageImports } = require("../../helpers");
+
+// https://github.com/patternfly/patternfly-react/pull/8737
+module.exports = {
+  meta: {},
+  create: function (context) {
+    const imports = getPackageImports(context, "@patternfly/react-core").filter(
```
thatblindgeye, line 7: Since we're only looking for a single component, we could do a `.find` instead of filter here:

```suggestion
    const emptyStateIconImport = getPackageImports(context, "@patternfly/react-core").find(
```

packages/eslint-plugin-pf-codemods/lib/rules/v5/emptyStateIcon-icon-required.js
```diff
@@ -0,0 +1,30 @@
...
+      (specifier) => "EmptyStateIcon" === specifier.imported.name
+    );
+    return imports.length === 0
+      ? {}
+      : {
+          JSXOpeningElement(node) {
+            if (imports.find((imp) => imp.local.name === node.name.name)) {
```
thatblindgeye, line 14: If we implemented the above comment, we also could avoid having to `.map` here and just check if the import `local.name === node.name.name`

thatblindgeye (changes requested):

packages/eslint-plugin-pf-codemods/lib/rules/v5/emptyStateIcon-icon-required.js
adamviktora: Good point, it would be confusing to say that "icon is required", when the fixer in the other rule converts the "component" into "icon".

adamviktora: Also, I have a question, why do we have to check that `.name` attribute exists on the `JSXOpeningElement`? (`a.name?.name`). I can't think of any use case where the JSXAttribute does not have a name, are there some special cases?

gitdallas (approved):

thatblindgeye (approved):

PR #346, opened 2023-03-24
chore(EmptyState): EmptyState API changes

### LINKED ISSUE #275: Empty State - Empty API changes from API clean up
Follow up to breaking change issue https://github.com/patternfly/patternfly-react/issues/8555

Placeholder. The Empty state Component will be cleaned up.  Impacted APIs should have code modes.  We will know more details one the PR is up.

Required actions:
1. Build codemod
2. Build test

adamviktora:
Closes #275 

emptyState-add-footer and emptyState-add-header rules produce only warnings, as these changes are not breaking, however fixer is still provided to follow new EmptyState API conventions


gitdallas:
Should we combine these? At least the header/footer as they use a lot of the same logic and hinge on the same elements.

packages/eslint-plugin-pf-codemods/lib/rules/v5/emptyState-rename-components.js
```diff
@@ -0,0 +1,10 @@
+const { renameComponents } = require("../../helpers");
+
+// https://github.com/patternfly/patternfly-react/pull/8737
+module.exports = {
+  meta: { fixable: "code" },
+  create: renameComponents({
+    "EmptyStatePrimary": "EmptyStateActions",
+    "EmptyStateSecondaryActions": "EmptyStateActions",
+  }),
+};
```
thatblindgeye, line 10: Since EmptyStatePrimary and EmptyStateSecondaryActions are being replaced completely, we should try to also remove their imports as part of the fix.

thatblindgeye (changes requested):
Some comments below. Would also agree with ☝🏼 re: combining the header and footer codemods if we can.

packages/eslint-plugin-pf-codemods/lib/rules/v5/emptyState-rename-components.js
adamviktora: @thatblindgeye @gitdallas 

So I have done this fix and I found out there are some issues related to how the `renameComponents` helper works. I also have some questions about alias imports.

1) Is there ever a case where the components are not replaced completely? Maybe let's say that some component becomes deprecated and it gets a new name. What the fixer inside `renameComponents` does, is that it **adds** new import (does not replace the old one), but on this contrary it renames all JSX elements. 

(now I see there is the `data-codemods` attribute which if present, don't rename the component, so I guess it makes sense not to replace the import)

2) If the component is aliased (e.g. `EmptyStatePrimary as ESP`), `renameComponents` helper will not rename the component. This would be fine, but only in the case, if the import would be **replaced** correctly -- so the import would look like `EmptyStateActions as ESP`. Instead it only adds `EmptyStateActions` import (not aliased), does not rename the JSXElements and keeps the import of the old component - so consumer would have to solve this fix manually.

3) In my specific case, when we are renaming both `EmptyStatePrimary` and `EmptyStateSecondaryActions` to the same component `EmptyStateActions`, if both components have aliased imports, let's say:
```
EmptyStatePrimary as Primary,
EmptyStateSecondaryActions as Secondary,
```
can we, after renaming, keep the EmptyStateActions import twice like this?
```
EmptyStateActions as Primary,
EmptyStateActions as Secondary,
```
Compilation seems to works fine.

4) When I try using alias import, I am getting this ReferenceError, do you know what is wrong?

Screenshot 2023-03-29 at 13 29 27

What do you think? In my opinion the `renameComponents` should be improved. Right now when I am using it, I have to handle removing imports separately and cannot handle the aliased imports. So I will end up not using it. Maybe I am missing something with the `data-codemods` attribute.

Well it was a long comment, I hope it all makes sense :D

thatblindgeye:
>Is there ever a case where the components are not replaced completely? Maybe let's say that some component becomes deprecated and it gets a new name. What the fixer inside renameComponents does, is that it adds new import (does not replace the old one), but on this contrary it renames all JSX elements.

Off the top of my head, the only cases I can think of are 1) our component deprecation where the components are being moved to a deprecated directory (and possibly a newer implementation is being promoted like Dropdown and Select), but we're only updating the import paths rather than replacing anything, or 2) We introduced a new component that can be used instead of/in addition to (similar to the EmptyStateFooter's I guess). 

>If the component is aliased (e.g. EmptyStatePrimary as ESP), renameComponents helper will not rename the component. This would be fine, but only in the case, if the import would be replaced correctly -- so the import would look like EmptyStateActions as ESP. Instead it only adds EmptyStateActions import (not aliased), does not rename the JSXElements and keeps the import of the old component - so consumer would have to solve this fix manually.

We *could* rename the import in a way that the alias remains (e.g. `EmptyStatePrimary as ESP` -> `EmptyStateActions as ESP`... Not sure if that makes more sense over just replacing the import and JSX element name completely with EmptyStateActions, though... Personally I might lean towards completely replacing an aliased import with just `EmptyStateActions`, even if there's a case like your 3rd point (Primary and Secondary both aliased originally -> after the fix only a single EmptyStateActions is imported rather than importing/aliasing it twice), but curious what Dallas thinks.

>When I try using alias import, I am getting this ReferenceError, do you know what is wrong?

My assumption is that the actual Markdown file for the component examples needs the aliased import towards the top of the file as well (similar to how the EmptyState.md file has `import CubesIcon from...` at the top). I've had an error when deprecating Dropdown stuff in react because an actual example tsx file was importing/aliasing a component, but the MD file didn't have the aliased import.

>What do you think? In my opinion the renameComponents should be improved. Right now when I am using it, I have to handle removing imports separately and cannot handle the aliased imports. So I will end up not using it. Maybe I am missing something with the data-codemods attribute.

I would agree that the helper could use some improvement. I really can't imagine a case where we would want to rename a component, but *only* at the JSX element level. To me "rename component" means we're completely getting rid of the old name, and it won't actually be importable (able to be imported? 😆 ) anymore.

That said, the lib/rules/v4 directory does have a `no-unused-imports` codemod... @gitdallas would it be easier to just bring that over to v5 as well? I wonder if that might be partly why the renameComponents helper only adds the new component name to the imports list rather than replacing the old name.

>Well it was a long comment, I hope it all makes sense :D

You definitely brought up some good points!

gitdallas:
I think @thatblindgeye responded well. Some thoughts..

In regard to alias, on the import specifier the imported.name will always show what's being exported while the local.name will show the alias if it exists (otherwise the same as imported.name) so when you look to compare JSXElement names, you always want to compare with local name. In the most recent commits you can see a helper for moving imports and updating jsxelement names, those might be helpful to look at.

I agree with Eric in that I don't think it would make sense to import the same thing twice with different aliases. For each, just see if the import.name exists (create if not), you could keep the alias (local name) if it exists already and use that to name both the jsxelement names if you want.

Yes for examples you have to have the import in the parent MD file for the examples to have access to them.

And yes let's look at the no-unused-imports... not a huge concern as a normal lint afterwards should take care of it though.

gitdallas:
Hi @adamviktora - i just tested this
image

and it ended up like this with the fixers:

image

gitdallas (changes requested):
Hi @adamviktora - i just tested this
image

and it ended up like this with the fixers:

image

We want reduce to a single import, even if there are aliases, and the JSXElement would need to keep track of which alias/localname we are using. It would be fine if you just use one of the aliases in the new import and update the JSXElements accordingly... or just removing the alias all together.

So the above should fix to something like..
image
or
image

adamviktora:
> Hi @adamviktora - i just tested this image
> 
> and it ended up like this with the fixers:
> 
> image
> 
> We want reduce to a single import, even if there are aliases, and the JSXElement would need to keep track of which alias/localname we are using. It would be fine if you just use one of the aliases in the new import and update the JSXElements accordingly... or just removing the alias all together.
> 
> So the above should fix to something like.. image or image
> 
> also please combine the header/footer rules

Yeah, I know, I did that commit before I read your replies, so that version does not remove aliases. Working on a new version now and will combine header/footer right after that :)

adamviktora:
@thatblindgeye @gitdallas Thank you both for great answers and suggestions! Should be all fixed and working now, I ended up using some core parts of the `no-unused-imports` rule from v4.

About the `renameComponents` helper, should I create a new issue for that one, which would include making it work on aliased components and removing the unused imports?

packages/eslint-plugin-pf-codemods/lib/rules/v5/emptyState-rename-components.js
```diff
@@ -0,0 +1,86 @@
...
+    const importNames = Object.fromEntries(
+      imports.map((imp) => [imp.local.name, imp.imported.name])
+    );
+
+    const newName = "EmptyStateActions";
+
+    return imports.length == 0
+      ? {}
+      : {
+          ImportDeclaration(node) {
```
thatblindgeye, line 23: I don't think I'd block over it since we could do a followup if needed, but I'm wondering if we could simplify the fixer for the ImportDeclaration so that we don't have to have a test file that needs multiple runs to remove components then add the new one. I can try testing something out quickly tomorrow morning.

thatblindgeye (changes requested):
This is looking great with the updates! Had 2 changes below and a general comment that right now isn't blocking/I want to look into a little tomorrow

packages/eslint-plugin-pf-codemods/lib/rules/v5/emptyState-rename-components.js
wise-king-sullyman: Followup made here: #369

wise-king-sullyman (approved):

thatblindgeye (approved):

gitdallas (approved):

PR #348, opened 2023-03-24
chore(FormGroup, FormHelperText): remove props

kmcfaul:
Removes props and warns about using `HelperText`, `HelperTextItem`, and `FormHelperText`. Had initially tried to insert the replacement node but there are many different translations that could be required based on which properties are present and what the value of the properties are (especially how the `validated` prop would translate into the new structure). I can fiddle with it more if we want to try to build the replacement HelperText block.

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/formgroup-remove-helpertextProps.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/formgroup-remove-helpertextProps.js
@@ -1,19 +1,29 @@
-const { renameProp } = require('../../helpers');
+const { getPackageImports } = require("../../helpers");
 
 // https://github.com/patternfly/patternfly-react/pull/8810
 module.exports = {
-  meta: { fixable: 'code' },
-  create: renameProp(
-    'FormGroup',
-    { 
-      helperText: '',
-      helperTextInvalid: '',
-      validated: '',
-      helperTextIcon: '',
-      helperTextInvalidIcon: '',
-      isHelperTextBeforeField: '' 
-    },
-    (node, attribute) =>
-      `${attribute.name.name} prop for ${node.name.name} has been removed. Use FormHelperText, HelperText, and HelperTextItem directly inside children.`
-  ),
+  meta: {},
+  create: function (context) {
+    const formGroupImport =  getPackageImports(context, '@patternfly/react-core')
+    .filter(specifier => specifier.imported.name == 'FormGroup');
+
+    return formGroupImport.length === 0 ? {} : {
+      JSXOpeningElement(node) {
+        if (formGroupImport.map(imp => imp.local.name).includes(node.name.name)) {
+          const helperProps = node.attributes.filter(a => {
+            const attr = a.name?.name;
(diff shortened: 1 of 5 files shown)


gitdallas:
I'm curious if we should offer a fixer if we aren't going to do the replacement, I think as a product dev I'd rather have it either do the entire fix or leave it alone especially in this case where having the values in the props would help in moving it manually to a child element.

Thoughts @thatblindgeye @jenny-s51 @wise-king-sullyman ?

thatblindgeye:
>I'm curious if we should offer a fixer if we aren't going to do the replacement, I think as a product dev I'd rather have it either do the entire fix or leave it alone especially in this case where having the values in the props would help in moving it manually to a child element.

Yeah I'd agree. A replacement fix would be nice to have, but I don't think I'd block over it just because of what Katie said re: "there are many different translations that could be required..."

kmcfaul:
@gitdallas Should this register as a warning if it won't have a fixer then? Or should it still be an error because it would error the user build if the props remain?

gitdallas:
@kmcfaul - i agree it would make sense to have it be an error if it would gaurantee an error in their code.

packages/eslint-plugin-pf-codemods/lib/rules/v5/formgroup-remove-helpertextProps.js
```diff
@@ -0,0 +1,29 @@
...
+  create: function (context) {
+    const formGroupImport =  getPackageImports(context, '@patternfly/react-core')
+    .filter(specifier => specifier.imported.name == 'FormGroup');
+
+    return formGroupImport.length === 0 ? {} : {
+      JSXOpeningElement(node) {
+        if (formGroupImport.map(imp => imp.local.name).includes(node.name.name)) {
+          const helperProps = node.attributes.filter(a => {
+            const attr = a.name?.name;
+            return attr === 'helperText' || attr === 'helperTextInvalid' || attr === 'validated' || attr === 'helperTextIcon' || attr === 'helperTextInvalidIcon' || attr === 'isHelperTextBeforeField'
```
thatblindgeye, line 15: Nit: could do an array of all the strings here and check if it `.includes(a.name.name)` instead

packages/pf-codemods/README.md
```diff
@@ -740,6 +740,28 @@ Out:
 <FileUpload  />
 ```
 
+### formgroup-remove-helpertextProps [(#8810)](https://github.com/patternfly/patternfly-react/pull/8810)
+
+We've removed the helper text related props from `FormGroup`. This rule will remove the `helperText`, `helperTextInvalid`, `validated`, `helperTextIcon`, `helperTextInvalidIcon`, and `isHelperTextBeforeField` props if present. The `FormHelperText`, `HelperText`, and `HelperTextItem` components should now be used directly as part of `children` instead of these props.
```
thatblindgeye, line 745: This just needs to be updated since we're not removing them anymore. Should also add, "This rule will throw an error, but will not make any changes." at the end.

thatblindgeye (changes requested):

gitdallas (approved):

wise-king-sullyman (approved):

PR #349, opened 2023-03-24
chore(Menu): rename MenuInput

kmcfaul:
To `MenuSearchInput` and add `MenuSearch` as a wrapper.

Codemod for https://github.com/patternfly/patternfly-react/issues/8187

### DIFF
--- a/packages/eslint-plugin-pf-codemods/test/rules/v5/menu-search-rename.js
+++ b/packages/eslint-plugin-pf-codemods/test/rules/v5/menu-search-rename.js
@@ -1,0 +1,42 @@
+const ruleTester = require("../../ruletester");
+const rule = require("../../../lib/rules/v5/menu-search-rename");
+
+ruleTester.run("menu-search-rename", rule, {
+  valid: [
+    {
+      code: `import { MenuSearchInput, MenuSearch } from '@patternfly/react-core'; <MenuSearch><MenuSearchInput /></MenuSearch>`,
+    },
+    // No @patternfly/react-core import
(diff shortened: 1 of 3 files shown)


gitdallas (approved):
nice. at first i thought there would be an error with non self-closing as you were using `JSXOpeningElement` rather than `JSXElement`, but i see how you are traversing. lgtm

packages/eslint-plugin-pf-codemods/test/rules/v5/menu-search-rename.js
```diff
@@ -0,0 +1,42 @@
...
+    {
+      code: `import { MenuSearchInput, MenuSearch } from '@patternfly/react-core'; <MenuSearch><MenuSearchInput /></MenuSearch>`,
+    },
+    // No @patternfly/react-core import
+    { code: `<MenuInput />` },
+  ],
+  invalid: [
+    {
+      code: `import { MenuInput } from '@patternfly/react-core'; <MenuInput><div /></MenuInput>`,
+      output: `import { MenuInput, MenuSearch, MenuSearchInput } from '@patternfly/react-core'; <MenuSearch><MenuSearchInput><div /></MenuSearchInput></MenuSearch>`,
```
thatblindgeye, line 15: Only thing is that I feel like we should replace MenuInput in the output rather than keeping it and only adding MenuSearchInput.

thatblindgeye (commented):
Just had one comment below

thatblindgeye (approved):

PR #351, opened 2023-03-24
chore(Datepicker): warn helperText

kmcfaul:
Warns that helperText can now accept HelperText components.

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/datepicker-warn-helperText.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/datepicker-warn-helperText.js
@@ -1,0 +1,30 @@
+const { getPackageImports } = require("../../helpers");
+
+  // https://github.com/patternfly/patternfly-react/pull/8833
+  module.exports = {
+    meta: {},
+    create: function (context) {
+      const imports = getPackageImports(
+        context,
+        "@patternfly/react-core"
+      ).filter((specifier) => specifier.imported.name === "DatePicker");
+
+      return imports.length === 0
+        ? {}
+        : {
+            JSXOpeningElement(node) {
+              if (
+                imports
+                  .map((imp) => imp.local.name)
+                  .includes(node.name.name)
+                  && node.attributes.find(a => a.name.name === 'helperText')
+              ) {
+                context.report({
+                  node,
+                  message: 'The helperText property now accepts the <HelperText> component.',
+                });
+              }
+            },
+          };
+    },
+  };
--- a/packages/pf-codemods/README.md
+++ b/packages/pf-codemods/README.md
@@ -579,6 +579,10 @@ function onChange2(_event, checked) {};
 
 The default value of the `appendTo` prop on DatePicker has been updated, which may cause markup changes that require updating selectors in tests. This rule will raise a warning, but will not make any changes.
 
+### datepicker-warn-helperText [(#8833)](https://github.com/patternfly/patternfly-react/pull/8833)
+
+The helperText property now accepts the <HelperText> component.
+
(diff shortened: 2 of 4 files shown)


packages/eslint-plugin-pf-codemods/lib/rules/v5/datepicker-warn-helperText.js
```diff
@@ -0,0 +1,30 @@
+const { getPackageImports } = require("../../helpers");
+
+  // https://github.com/patternfly/patternfly-react/pull/8833
+  module.exports = {
+    meta: {},
+    create: function (context) {
+      const imports = getPackageImports(
+        context,
+        "@patternfly/react-core"
+      ).filter((specifier) => specifier.imported.name === "DatePicker");
```
gitdallas, line 10: no biggie but as you are just getting a single thing here you can use `find` and then not have to map it out below

gitdallas (approved):
lgtm, small nit below

packages/eslint-plugin-pf-codemods/lib/rules/v5/datepicker-warn-helperText.js
```diff
@@ -0,0 +1,30 @@
...
+            JSXOpeningElement(node) {
+              if (
+                imports
+                  .map((imp) => imp.local.name)
+                  .includes(node.name.name)
+                  && node.attributes.find(a => a.name.name === 'helperText')
+              ) {
+                context.report({
+                  node,
+                  message: 'The helperText property now accepts the <HelperText> component.',
```
thatblindgeye, line 24: Based on the latest updates to the associated react PR, this message will need to be updated since now whatever is passed to helperText is automatically wrapped in HelperText components.

packages/pf-codemods/README.md
```diff
@@ -579,6 +579,10 @@ function onChange2(_event, checked) {};
 
 The default value of the `appendTo` prop on DatePicker has been updated, which may cause markup changes that require updating selectors in tests. This rule will raise a warning, but will not make any changes.
 
+### datepicker-warn-helperText [(#8833)](https://github.com/patternfly/patternfly-react/pull/8833)
+
+The helperText property now accepts the <HelperText> component.
```
thatblindgeye, line 584: Same as above comment, this will probably need to be updated.

thatblindgeye (commented):
In addition to below, might make sense to add Timepicker into this codemod since it's the same PR, but not a blocker if not.

packages/eslint-plugin-pf-codemods/lib/rules/v5/datepicker-warn-helperText.js
kmcfaul: The Datepicker PR now does not automatically wrap the prop. I will update the message to say "expects" instead.

packages/pf-codemods/README.md
```diff
@@ -579,6 +579,10 @@ function onChange2(_event, checked) {};
 
 The default value of the `appendTo` prop on DatePicker has been updated, which may cause markup changes that require updating selectors in tests. This rule will raise a warning, but will not make any changes.
 
+### datetimepicker-warn-helperText [(#8833)](https://github.com/patternfly/patternfly-react/pull/8833)
+
+The helperText property of `DatePicker` now accepts the <HelperText> component, and `TimePicker` now uses a <HelperText> component internally.
```
thatblindgeye, line 584:
```suggestion
The helperText property of `DatePicker` now expects the <HelperText> component, and `TimePicker` now uses a <HelperText> component internally.

This rule will raise a warning, but will not make any code changes
```

thatblindgeye (changes requested):
Quick change below

PR #352, opened 2023-03-24
chore(Table): refine ActionsColumn warning

kmcfaul:
Now will issue a warning, and only if both the Table import and ActionsColumn/DropdownToggle are also imported.

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/table-warn-actionsColumn.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/table-warn-actionsColumn.js
@@ -29,7 +29,7 @@ module.exports = {
         if (tableImport && (dropdownToggleImport || actionsColumnImport)) {
           context.report({
             node,
-            message: "The ActionsColumn now uses the Next version of Dropdown. The action toggle should now pass a MenuToggle rather than a DropdownToggle, and direction and position properties are now passed under the ActionsColumn new popperProps property.",
+            message: "The ActionsColumn within Table now uses our new implementation of Dropdown. The action toggle should now pass a MenuToggle rather than the deprecated DropdownToggle, and direction and position properties are now passed under the ActionsColumn new popperProps property.",
           });
         }
       },
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/table-warn-actionsColumn.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/table-warn-actionsColumn.js
@@ -6,8 +6,10 @@ module.exports = {
     return {
       ImportDeclaration(node) {
         const coreImports = getPackageImports(context, '@patternfly/react-core');
+        const deprecatedImports = getPackageImports(context, '@patternfly/react-core/deprecated');
         const tableImports = getPackageImports(context, '@patternfly/react-table');
-        const imports = [...coreImports, ...tableImports];
+        const deprecatedTableImports = getPackageImports(context, '@patternfly/react-table/deprecated');
+        const imports = [...coreImports, ...tableImports, ...deprecatedImports, ...deprecatedTableImports];
 
         const tableImport = imports.find(
           (specifier) =>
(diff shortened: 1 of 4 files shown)


gitdallas:
I wonder if we should continue using `importDeclaration` so we can point the codemod report to the line number of the import declaration.

also, @thatblindgeye - curious how this will work in relation to https://github.com/patternfly/pf-codemods/pull/344

on first pass without fix... their code will likely be pointing to `DropdownToggle` in `react-core` which in v4 was what will be `react-core/deprecated` in v5... will trigger both these codemods.. fine? Then after running a fix, it will point to "deprecated" and this one shouldn't be triggered (at least by the table+dropdownToggle)

thatblindgeye:
>on first pass without fix... their code will likely be pointing to DropdownToggle in react-core which in v4 was what will be react-core/deprecated in v5... will trigger both these codemods.. fine? Then after running a fix, it will point to "deprecated" and this one shouldn't be triggered (at least by the table+dropdownToggle)

That sounds accurate to me. Whichever one (this or 344) gets in last we can double check that though in the test.tsx file and tweak things from there if needed.

gitdallas:
@thatblindgeye @kmcfaul  an update on how it looks with the dropdown deprecation:

image

running the fix will clear the codemod errors/warnings look like this:

image

should this issue also (only?) check for the `/deprecated` version?

thatblindgeye:
> should this issue also (only?) check for dropdown from the /deprecated import path?

Maybe...? If we're not trying/expecting for the fix flag to fix everything with a single pass, then it may not be a bad idea to only target the deprecated path.

kmcfaul:
Wouldn't that mean the order that the codemods run in will matter then? Because the codemod to update the deprecated path would have to run its fix before this codemod would detect the conditions to warn the user about the mismatch. I could update the codemod to look in both directories maybe, or am I overthinking this?

packages/eslint-plugin-pf-codemods/lib/rules/v5/table-warn-actionsColumn.js
```diff
@@ -1,18 +1,35 @@
...
+        const actionsColumnImport = imports.find(
+          (specifier) =>
+            (specifier.imported.name === "ActionsColumn")
+        );
+
+        if (tableImport && (dropdownToggleImport || actionsColumnImport)) {
           context.report({
             node,
-            message: "The ActionsColumn now uses the Next version of Dropdown. The action toggle should now pass a MenuToggle rather than a DropdownToggle, and direction and position properties are now passed under the new popperProps property.",
+            message: "The ActionsColumn now uses the Next version of Dropdown. The action toggle should now pass a MenuToggle rather than a DropdownToggle, and direction and position properties are now passed under the ActionsColumn new popperProps property.",
```
thatblindgeye, line 32: Latest update should work.. I think there's pros and cons to targeting only the deprecated path for DropdownToggle vs both react-core and deprecated like the latest update does.

Updating the warning verbiage may help:

```suggestion
            message: "The ActionsColumn within Table now uses our new implementation of Dropdown. The action toggle should now pass a MenuToggle rather than the deprecated DropdownToggle, and direction and position properties are now passed under the ActionsColumn new popperProps property.",
```

Maybe make the "within Table" dynamic so that it uses whatever name is found from tableImport, wdyt?

Could we also update the README for this codemod so that the opening sentence states, "...has been updated to use our new implementation of Dropdown"

gitdallas:
> Wouldn't that mean the order that the codemods run in will matter then? 

we may run into this situation.. v4 codemods have a similar issue in `card-rename-components` where the renames were as such:

```
const renames = {
  'CardHeader': 'CardTitle',
  'CardHead': 'CardHeader',
  'CardHeadMain': 'CardHeaderMain'
};
```

so CardHeader exists both before and after... but differently. They solved the issue by adding a `data-codemods` attribute. We may have to consider this at some point if we have situations where we can't differentiate otherwise.

gitdallas (approved):

wise-king-sullyman (approved):

Issue #359, opened 2023-03-30
lint error thrown when there is no space before callback arg (not a codemod)


Dominik-Petrik:
When there is no space between comma and argument in param callback, trying to swap the arguments results in lint error.
`
<Component onChange={(value,event) => handleChange(value, event)}/>
`

PR #361, opened 2023-03-30
fix error thrown when space missing before callback arg

Dominik-Petrik:
closes: #359

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/helpers.js
+++ b/packages/eslint-plugin-pf-codemods/lib/helpers.js
@@ -552,10 +552,10 @@ function addCallbackParam(componentsArray, propMap) {
                       const createRemoveCurrentParamUseFix = (
                         currentUseOfNewParam
                       ) => {
-                        let tokenBeforeCurrentUse = context.getTokenBefore(currentUseOfNewParam)
+                        const tokenBeforeCurrentUse = context.getTokenBefore(currentUseOfNewParam)
                         const targetRange = [
-                          tokenBeforeCurrentUse.value === "," ? tokenBeforeCurrentUse.range[0] : tokenBeforeCurrentUse.range[1] + 1,
-                          currentUseOfNewParam.range[1],
+                          tokenBeforeCurrentUse.range[0],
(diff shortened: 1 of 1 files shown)


packages/eslint-plugin-pf-codemods/lib/helpers.js
```diff
@@ -552,8 +552,9 @@ function addCallbackParam(componentsArray, propMap) {
                       const createRemoveCurrentParamUseFix = (
                         currentUseOfNewParam
                       ) => {
+                        let tokenBeforeCurrentUse = context.getTokenBefore(currentUseOfNewParam)
                         const targetRange = [
-                          currentUseOfNewParam.range[0] - 2,
+                          tokenBeforeCurrentUse.value === "," ? tokenBeforeCurrentUse.range[0] : tokenBeforeCurrentUse.range[1] + 1,
```
wise-king-sullyman, line 557: Will the token before ever not be `,`?

packages/eslint-plugin-pf-codemods/lib/helpers.js
```diff
@@ -552,8 +552,9 @@ function addCallbackParam(componentsArray, propMap) {
                       const createRemoveCurrentParamUseFix = (
                         currentUseOfNewParam
                       ) => {
+                        let tokenBeforeCurrentUse = context.getTokenBefore(currentUseOfNewParam)
```
thatblindgeye, line 555: Nit: can we make this a const instead of let?

thatblindgeye (changes requested):

packages/eslint-plugin-pf-codemods/lib/helpers.js
Dominik-Petrik: Yep, you're right, thanks.

packages/eslint-plugin-pf-codemods/lib/helpers.js
Dominik-Petrik: It probably won't but I wasn't sure.

thatblindgeye (approved):

PR #366, opened 2023-03-30
feat(DataList): update selectableRow.onChange to onSelectableRowChange

nicolethoen:
Closes: https://github.com/patternfly/pf-codemods/issues/268

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/datalist-remove-selectableRow.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/datalist-remove-selectableRow.js
@@ -14,7 +14,7 @@ module.exports = {
           if (selectableRow) {
             context.report({
               node,
-              message: `DataList's selectableRow.onChange property has been replaced with onSelectableRowChange. The order of the params in the callback has also been updated so that the event param is first.`,
+              message: `DataList's selectableRow property has been replaced with onSelectableRowChange. The order of the params in the callback has also been updated so that the event param is first.`,
             });
           }
         }
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/datalist-remove-selectableRow.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/datalist-remove-selectableRow.js
@@ -1,0 +1,25 @@
+const { renameProp, getPackageImports} = require('../../helpers');
+
+//https://github.com/patternfly/patternfly-react/pull/8827
+module.exports = {
+  meta: {},
+  create: function (context) {
+    const dataListImport =  getPackageImports(context, '@patternfly/react-core')
+      .filter(specifier => specifier.imported.name == 'DataList');
+
+    return dataListImport.length === 0 ? {} : {
+      JSXOpeningElement(node) {
+        if (dataListImport.map(imp => imp.local.name).includes(node.name.name)) {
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin-pf-codemods/lib/rules/v5/datalist-remove-selectableRow.js
```diff
@@ -0,0 +1,25 @@
...
+      .filter(specifier => specifier.imported.name == 'DataList');
+
+    return dataListImport.length === 0 ? {} : {
+      JSXOpeningElement(node) {
+        if (dataListImport.map(imp => imp.local.name).includes(node.name.name)) {
+          const selectableRow = node.attributes.find(a => a.name?.name === 'selectableRow');
+          if (selectableRow) {
+            context.report({
+              node,
+              message: `DataList's selectableRow.onChange property has been replaced with onSelectableRowChange. The order of the params in the callback has also been updated so that the event param is first.`,
```
thatblindgeye, line 17: Nit: Would saying "DataList's selectableRow prop has been replaced with the onSelectableRowChange callback" be more clear?

packages/pf-codemods/README.md
```diff
@@ -514,6 +514,38 @@ Out:
 <DataList      />
 ```
 
+### datalist-remove-selectableRow [(#8827)](https://github.com/patternfly/patternfly-react/pull/8827)
+
+We've removed the selectableRow.onChange property and replaced it with onSelectableRowChange. The order of the params in the callback has also been updated so that the event param is first. This will require
+
+#### Example of manual change needed
```
thatblindgeye, line 521: 🔥 I like the idea of this. This one *might* be able to have a fix applied as a followup, but for right now this is a good fallback and might be good to look at our other codemods that don't provide a fix.

thatblindgeye (changes requested):

packages/eslint-plugin-pf-codemods/lib/rules/v5/datalist-remove-selectableRow.js
nicolethoen: maybe - I wondered how to succinctly convey that the prop name AND type has updated - as well as the param order in the callback. but I may not be giving enough credit to the consumer to be able to figure it out.

packages/pf-codemods/README.md
```diff
@@ -516,7 +516,9 @@ Out:
 
 ### datalist-remove-selectableRow [(#8827)](https://github.com/patternfly/patternfly-react/pull/8827)
 
-We've removed the selectableRow.onChange property and replaced it with onSelectableRowChange. The order of the params in the callback has also been updated so that the event param is first. This will require
+We've removed the selectableRow property and replaced it with onSelectableRowChange. The value of the selectableRow's onChange field
+was a callback, which can now be directly passed to the onSelectableRowChange prop. However, it's worth nothing that the order of the params
```
thatblindgeye, line 520:
```suggestion
was a callback, which can now be directly passed to the onSelectableRowChange prop. However, it's worth noting that the order of the params 
```

thatblindgeye (commented):
Looks good! Quick typo fix below and the message in the lib/rules/v5 file just needs to be updated to match the error message from the test file

thatblindgeye (approved):

gitdallas (approved):

wise-king-sullyman (approved):
Made a followup here to track investigating a fixer #368

PR #367, opened 2023-03-30
improve generators

wise-king-sullyman:
Refactors the generators to get rid of duplicated code, make it easier to create new generators, and have them all use a single interactive CLI.

Also adds the ability for generators to update the readme and test.tsx files, including placing new content where it should be alphabetically when needed.

------------------------------------------------------------

Recommend reviewing commit by commit.
- chore(generators): Improve generator capability and usability
- chore(readme, test.tsx): Fix alphabetization errors
- chore(generators): Remove old generators
- chore(generators): Fix import updating bug in test.tsx generator
- chore(test.tsx): Remove trailing whitespaces in imports

------------------------------------------------------------

To test:
1. checkout branch
2. install deps
3. run `yarn generate`
4. follow prompts from interactive CLI

While you're doing that please note anything you find confusing. I plan to update the readme to document how to use it, but I'm not sure yet what information that needs to include.


gitdallas (approved):
Lgtm although i didn't look super close, but figured it won't hurt anything and might as well get it in so people can really start trying it out

gitdallas (approved):

generators/update-readme.js
```diff
@@ -0,0 +1,95 @@
...
+Out:
+
+\`\`\`jsx
+${outputExamples}
+\`\`\`
+
+`;
+
+  splitRules.push(readMeAddition);
+  const newRules = splitRules.sort().join("\n### ");
```
thatblindgeye, line 38: Right now this results in `datalist-remove-props` coming after `dataListCheck-warn-updated-callback`, and an import of `kebabToggle` coming after `KEY_CODES`. Would we want to ignore case when sorting, or do you think it makes more sense to take lettercase into account?

Not really a blocker, but I did notice it.

thatblindgeye (commented):
Looks great! One file comment below. The plan is still to do a followup with some instructions/description for using the generator right?

generators/update-readme.js
thatblindgeye: Let me know if you want to try implementing it, otherwise I'm good merging in as-is and we can do followups for this.

thatblindgeye: Wiki entry could work, or maybe a separate README somewhere

thatblindgeye (approved):

generators/update-readme.js
wise-king-sullyman: Ah yeah, the contributing readme would make sense

PR #372, opened 2023-04-04
fixing table-warn-actionsColumn to only warn on pf table imports

gitdallas:
These warnings were popping up for *every* import if the scenario existed.. example in koku:
image


thatblindgeye (approved):
Looks good. Am now wondering though whether it makes more sense to have the warning occur on the Table import (where the warning would be originating from technically) or the DropdownToggle (what would be causing the warning/need updating).

wise-king-sullyman (approved):
I think I'd probably lean towards having it point to the DropdownToggle, but I wouldn't block over it either.

gitdallas:
@thatblindgeye @wise-king-sullyman it could also be the `ActionsColumn` that triggers this, and that comes from `react-table`

we could update the logic to check for `table` with `ActionsColumn` specifier and then check for `core` with `DropdownToggle` if you'd prefer.

Issue #377, opened 2023-04-05
addCallbackParam helper - fix reporting issues (not a codemod)


thatblindgeye:
- Should make sure the addCallbackParam helper takes into account the components that are passed to another component's props, e.g.:

    ```
    <ComponentA toggle={<ComponentB onToggle={...} />}
    ```

    is caught.

- should add `___Deprecated` to the components list passed to the helper (or replace with the deprecated names) as this may cause params to not be updated
- look into errors being reported even after the fix has been applied

gitdallas:
seen in koku-ui, if the import path is moved to deprecated before the param change happens, it will be missed.

![image](https://user-images.githubusercontent.com/5322142/230158645-3343f3e0-d644-44ae-991a-47aad6158c2e.png)

PR #383, opened 2023-04-06
fix(addCallBackParam): helper will now also check deprecated import src

gitdallas:
#377


wise-king-sullyman (approved):

thatblindgeye (approved):
Should either mark this as just "work towards" the linked issue rather than closing it, or open another issue for looking into components passed to props

gitdallas:
@thatblindgeye - i thought you tested that works fine and was probably just an issue with it being deprecate fixed first

thatblindgeye:
@gitdallas Ah yeah you're right. Brain is fried 😵 This looks good then!

jenny-s51 (approved):

PR #384, opened 2023-04-06
feat(FileUploadField): Add cb param change mods

### LINKED ISSUE #355: FileUpload - change param order for onChange so event is first and add event param to onTextChange of FileUploadField
This is a test of creating a codemod which assist with the breaking change.

_Required actions:_
1. Build codemod
2. Build test

wise-king-sullyman:
Closes #355 

Also the changes in the index files enable us to exclude certain rules by default and only run them when using the `--only` flag.

### DIFF
--- a/packages/eslint-plugin-pf-codemods/index.js
+++ b/packages/eslint-plugin-pf-codemods/index.js
@@ -1,4 +1,7 @@
-const createListOfRules = (version) => {
+// if you want your rule to only run when explicitly called for using the --only flag, add the rule name to the below array
+const betaRuleNames = ["fileUploadField-cb-param-updates"];
+
+const createListOfRules = (version, includeBeta = false) => {
   const rules = {};
   require("glob")
     .sync(
@@ -8,13 +11,18 @@ const createListOfRules = (version) => {
     )
     .forEach(function (file) {
       const ruleName = /.*\/([^.]+)/.exec(file)[1];
-      rules[ruleName] = require(`./lib/rules/v${version}/${ruleName}`);
+      const isBeta = betaRuleNames.includes(ruleName);
+
+      if (includeBeta === isBeta) {
+        rules[ruleName] = require(`./lib/rules/v${version}/${ruleName}`);
+      }
     });
   return rules;
 };
 
 const v5rules = createListOfRules("5");
 const v4rules = createListOfRules("4");
+const betaV5Rules = createListOfRules("5", true);
 
 // if you want a rule to have a severity that defaults to warning rather than error, add the rule name to the below array
 const warningRules = [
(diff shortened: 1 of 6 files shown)


packages/eslint-plugin-pf-codemods/index.js
```diff
@@ -1,4 +1,7 @@
-const createListOfRules = (version) => {
+// if you want your rule to only run when explicitly called for using the --only flag, add the rule name to the below array
```
thatblindgeye, line 1: Would be worth adding some verbiage to the contributing MD about this new array, as well as when you'd want to add to it.

thatblindgeye (commented):
Quick comment below. Is the intention to get this in and then make additional updates to this codemod as a followup? Mainly asking because there are some other things I'm noticing, but if we just want this in to test the new beta array out then I can hold off until a followup.

wise-king-sullyman:
I do plan to make some additional updates to the codemod after I actually use it and make the change in pf-react, but if you'd like to let your comments fly here you can. If they aren't major and will take a decent amount of time to fix I can reference the discussion here when I come back to it if nothing else.

packages/eslint-plugin-pf-codemods/lib/rules/v5/fileUploadField-cb-param-updates.js
```diff
@@ -0,0 +1,14 @@
...
+module.exports = {
+  meta: { fixable: "code" },
+  create: addCallbackParam(["FileUploadField"], {
+    onChange: {
+      defaultParamName: "_event",
+      previousParamIndex: 2,
+      otherMatchers: /^_?(ev\w*|e$)/,
+    },
+    onTextChange: '_event'
+  }),
```
thatblindgeye, line 13:
```suggestion
  create: addCallbackParam(["FileUploadField"], {
    onTextChange: '_event'
  }),
```

onChange was removed from FileUploadField, so we probably don't need to supply a fix for a prop that will just get removed.

thatblindgeye (changes requested):
Was partially the commented out PR links 😆 Otherwise another comment below

thatblindgeye (approved):

packages/eslint-plugin-pf-codemods/lib/rules/v5/fileUploadField-cb-param-updates.js
```diff
@@ -0,0 +1,7 @@
+const { addCallbackParam } = require("../../helpers");
+
+// https://github.com/patternfly/patternfly-react/pull/1234
```
gitdallas, line 3: 1234

packages/pf-codemods/README.md
```diff
@@ -986,6 +986,35 @@ Out:
 <FileUpload  />
 ```
 
+### fileUploadField-cb-param-updates [(#1234)](https://github.com/patternfly/patternfly-react/pull/1234)
```
gitdallas, line 989: 1234

gitdallas (changes requested):
probably don't mean to reference a pr from 2019

gitdallas (approved):

PR #386, opened 2023-04-06
feat(Table): update paths to reference deprecated directory

### LINKED ISSUE #330: Table - update path to deprecated directory
Follow up to breaking change issue https://github.com/patternfly/patternfly-react/issues/8005

_Required actions:_
1. Build codemod
2. Build test

jenny-s51:
Closes #330

### DIFF
--- a/packages/pf-codemods/README.md
+++ b/packages/pf-codemods/README.md
@@ -1855,6 +1855,35 @@ Out:
 <Spinner  />
 ```
 
+### table-update-deprecatedPath [(#8892)](https://github.com/patternfly/patternfly-react/pull/8892)
+
+We've deprecated the current implementation of Table. In order to continue using this deprecated implementation, the import path must be updated to our deprecated package and specifiers must be aliased. However, we suggest updating to our composable Table implementation.
+
+The following imports will be affected by this deprecation:
+
+- Table
+- TableBody
+- TableHeader
+- TableProps
+
+#### Examples
+
+In:
+
+```jsx
+import { Table, TableBody, TableHeader, TableProps } from "@patternfly/react-table";
+```
+
+Out:
+
+```jsx
+import {
+ Table as TableDeprecated,
+ TableBody as TableBodyDeprecated,
+ TableHeader as TableHeaderDeprecated,
+ TableProps as TablePropsDeprecated
+} from '@patternfly/react-table/deprecated';```
+
 ### table-warn-actionsColumn [(#8629)](https://github.com/patternfly/patternfly-react/pull/8629)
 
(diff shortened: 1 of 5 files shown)


packages/pf-codemods/README.md
```diff
@@ -1552,6 +1552,17 @@ Out:
...
+
+We've deprecated the current implementation of Table. In order to continue using this deprecated implementation, the import path must be updated to our deprecated package and specifiers must be aliased. However, we suggest updating to our composable Table implementation.
+
+The following imports will be affected by this deprecation:
+
+- Table
+- TableBody
+- TableHeader
+- TableProps
+
```
thatblindgeye, line 1565: Can we also include examples for the input/output?

thatblindgeye (commented):
I think this looks good, just a change noted below. The main thing is that this and the codemod that updates TableComposable to just Table will cause this codemod to run after the rename (if a consumer is running fixes more than once).

Wondering if we could update the logic to moveSpecifiers a little so that if other composable Table components are found in the import declaration to skip the import path rename. Since I can't imagine a consumer *only* importing Table without Th, Td, etc, which those wouldn't be imported from the new deprecated Table.

If we go that route we'd need to decide whether we'd want to include another argument to be passed in (a condition or array of components that tell the helper "if true or if these components are in the import declaration, don't rename the import path", or just add table logic internally.

jenny-s51 (commented):
Great points @thatblindgeye , thank you - I've updated #388 to add the `data-codemods="true"` attribute to `Table` which is the method that we were using in v4 to prevent multiple updates to codemods, thanks @gitdallas for this suggestion

packages/pf-codemods/README.md
```diff
@@ -1552,6 +1552,35 @@ Out:
...
+Out:
+
+```jsx
+import {
+ Table as TableDeprecated,
+ TableBody as TableBodyDeprecated,
+ TableHeader as TableHeaderDeprecated,
+ TableProps as TablePropsDeprecated
+} from '@patternfly/react-table/deprecated';```
```
wise-king-sullyman, line 1582: I think if the backticks aren't on a newline they'll cause some formatting issues.

wise-king-sullyman (commented):
One markdown formatting nit:

wise-king-sullyman (approved):

gitdallas (approved):

thatblindgeye (approved):
Just needs a rebase

jenny-s51 (commented):
@thatblindgeye rebased - should be all set to merge

wise-king-sullyman (approved):

PR #387, opened 2023-04-07
fix(EmptyState): codemod issues

adamviktora:
Closes: https://github.com/patternfly/pf-codemods/issues/376

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/emptyState-warn-change-structure.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/emptyState-warn-change-structure.js
@@ -17,8 +17,8 @@ module.exports = {
     const preFooterNames = [
       "EmptyStateBody",
       "EmptyStateHeader",
-      "EmptyStateIcon",
       "Title",
+      "EmptyStateIcon",
     ];
 
     const includesEmptyStateContent = imports.some((specifier) =>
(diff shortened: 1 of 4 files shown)


gitdallas (approved):
tested a couple scenarios and seems good

thatblindgeye (commented):
One issue I'm seeing now is that if there isn't anything that should render an EmptyStateHeader, a warning still gets raised and the fix adds the EmptyStateHeader import despite not using it.

adamviktora:
weird behaviour spotted: when `<EmptyState>` is not on a top level, it will not do the EmptyStateHeader fix

Before:
Screenshot 2023-04-14 at 9 25 50 (2)

After:
Screenshot 2023-04-14 at 9 27 10 (2)

But when having it on the top level, every EmptyState gets updated correctly

Before:
Screenshot 2023-04-14 at 9 29 16 (2)

After:
Screenshot 2023-04-14 at 9 30 02 (2)

Trying to see what is the problem right now

packages/eslint-plugin-pf-codemods/lib/rules/v5/emptyState-warn-change-structure.js
```diff
@@ -4,20 +4,59 @@ const { getPackageImports, ensureImports } = require("../../helpers");
...
+        }
+      };
+
+      searchForElements(baseElement, nameToGet);
+
+      return elementsArray;
+    };
+    const allElements = context
+      .getSourceCode()
+      .ast.body.filter((node) => node.expression?.openingElement);
```
adamviktora, line 37: here is a problem, because it only gets the nodes in the top level of ast.body

the getElements function looks like it recursively searches for elements inside, maybe some similar function like this is suitable for getting all JSXOpeningElements inside Program

adamviktora:
just found a way to get all JSXElements in a file (outside of the`return { JSXElement(node) {...} }`) portion of the rule, working on a fix, that will solve the imports issue

thatblindgeye (commented):
Awesome work on this! Could we add a test to check that a `<Title>` outside of `<EmptyState>` won't trigger the error (with all components imported from patternfly)? Something like:

```
    <>
      <EmptyState>
        <EmptyStateBody>
          Some other content.
        </EmptyStateBody>
      </EmptyState>
      <Title>Text</Title>
    </>
```

thatblindgeye (approved):

PR #388, opened 2023-04-07
feat(Table): rename TableComposable to Table

### LINKED ISSUE #385: Table - rename TableComposable
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/issues/8005

Rename TableComposable to Table following deprecation work

_Required actions:_
1. Build codemod
2. Build test

jenny-s51:
Closes #385

### DIFF
--- a/packages/eslint-plugin-pf-codemods/index.js
+++ b/packages/eslint-plugin-pf-codemods/index.js
@@ -30,7 +30,7 @@ const warningRules = [
 ];
 
 // rules that will run before other rules (move to deprecated?)
-const setupRules = [];
+const setupRules = ["table-update-deprecatedPath"];
 
 // rules that will run after other rules (cleanup imports?)
 const cleanupRules = ["no-unused-imports-v5"];
--- a/packages/eslint-plugin-pf-codemods/lib/helpers.js
+++ b/packages/eslint-plugin-pf-codemods/lib/helpers.js
@@ -10,11 +10,23 @@ function moveSpecifiers(
 ) {
   return function (context) {
     const importNames = importsToMove.map((nameToMove) => nameToMove.name);
-    const importSpecifiersToMove = getPackageImports(
+    const fromPackageImports = getPackageImports(
       context,
       fromPackage,
       importNames
     );
+    const allElements = getAllJSXElements(context);
+    const importSpecifiersToMove = fromPackageImports.filter((fromImport) => {
+      const foundElement = allElements.find(
+        (el) => el?.openingElement?.name?.name === fromImport.local?.name
+      );
+      const hasDataAttr = foundElement?.openingElement.attributes?.find(
+        (attr) => attr.name?.name === "data-codemods"
+      );
+
+      return (foundElement && !hasDataAttr) || !foundElement;
+    });
+
     if (!importSpecifiersToMove.length) return {};
 
     let modifiedToPackage = "";
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/table-rename-TableComposable.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/table-rename-TableComposable.js
@@ -1,0 +1,86 @@
+const {getPackageImports} = require("../../helpers");
+
+// https://github.com/patternfly/patternfly-react/pull/8892
+module.exports = {
+  meta: {fixable: "code"},
+  create: function (context) {
+    const imports = getPackageImports(context, "@patternfly/react-table");
+    const tableComposableImport = imports.find(
+      (imp) => imp.imported.name === "TableComposable"
+    );
+
+    return imports.length === 0 || !tableComposableImport
+      ? {}
+      : {
+          // update opening/closing elements
+          JSXIdentifier(node) {
+            const nodeName = node.name;
+            const importedNode = imports.find(
+              (imp) => imp.local.name === nodeName
+            );
+            const parentNode = node.parent;
+            const isOpeningTag = parentNode.type === "JSXOpeningElement";
+            const openingTagAttributes = isOpeningTag
+              ? parentNode.attributes
+              : parentNode.parent.openingElement?.attributes;
+            const hasDataAttr =
+              openingTagAttributes &&
+              openingTagAttributes.filter(
+                (attr) => attr.name?.name === "data-codemods"
+              ).length;
+            const isAliased =
+              importedNode?.local?.name !== importedNode?.imported?.name;
+            if (importedNode?.imported?.name === "Table" && hasDataAttr) {
+              // if data-codemods attribute, do nothing
+              return;
+            } else if (importedNode?.imported?.name === "TableComposable") {
+              const addDataAttr = (jsxStr) =>
+                jsxStr.replace(/(\s*\/?>)$/, ' data-codemods="true"$1');
+              const updateTagName = (node) => {
+                const tagName = context.getSourceCode().getText(node);
+
+                if (isAliased) {
+                  return tagName;
+                }
+                return tagName.replace("TableComposable", "Table");
+              };
+              const isOpeningTag = node.parent.type === "JSXOpeningElement";
+              const newOpeningParentTag = addDataAttr(
+                updateTagName(node.parent)
+              );
+              const newClosingTag = updateTagName(node);
+              (isOpeningTag || !isAliased) &&
+                context.report({
+                  node,
+                  message: `${nodeName} has been replaced with ${nodeName.replace(
+                    "TableComposable",
+                    "Table"
+                  )}`,
+                  fix(fixer) {
+                    return isOpeningTag
+                      ? fixer.replaceText(node.parent, newOpeningParentTag)
+                      : fixer.replaceText(node, newClosingTag);
+                  },
+                });
+            }
+          },
+          ImportSpecifier(node) {
+            const importedName = node.imported.name;
+            if (importedName === "TableComposable") {
+              const tableName = "Table";
+              const localName = node.local.name;
+              const isAliased = importedName !== localName;
+              const aliasText = isAliased ? ` as ${localName}` : "";
+              const newName = `${tableName}${aliasText}`;
+              context.report({
+                node,
+                message: `${importedName} has been replaced with ${tableName}`,
+                fix(fixer) {
+                  return fixer.replaceText(node, newName);
+                },
+              });
+            }
+          },
+        };
+  },
+};
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/table-rename-TableComposable.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/table-rename-TableComposable.js
@@ -23,11 +23,6 @@ module.exports = {
             const openingTagAttributes = isOpeningTag
               ? parentNode.attributes
               : parentNode.parent.openingElement?.attributes;
-            const hasDataAttr =
-              openingTagAttributes &&
-              openingTagAttributes.filter(
-                (attr) => attr.name?.name === "data-codemods"
-              ).length;
             const isAliased =
               importedNode?.local?.name !== importedNode?.imported?.name;
             if (importedNode?.imported?.name === "TableComposable") {
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/table-rename-TableComposable.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/table-rename-TableComposable.js
@@ -30,10 +30,7 @@ module.exports = {
               ).length;
             const isAliased =
               importedNode?.local?.name !== importedNode?.imported?.name;
-            if (importedNode?.imported?.name === "Table" && hasDataAttr) {
-              // if data-codemods attribute, do nothing
-              return;
-            } else if (importedNode?.imported?.name === "TableComposable") {
+            if (importedNode?.imported?.name === "TableComposable") {
               const addDataAttr = (jsxStr) =>
                 jsxStr.replace(/(\s*\/?>)$/, ' data-codemods="true"$1');
               const updateTagName = (node) => {
--- a/packages/pf-codemods/README.md
+++ b/packages/pf-codemods/README.md
@@ -2089,6 +2089,22 @@ import {
 } from '@patternfly/react-table/deprecated';
 ```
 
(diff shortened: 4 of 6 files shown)


thatblindgeye (commented):
This looks good! I'd want to hold off on approving and merging this until we can decide on a path in regards to this and the deprecation PR working together (as commented here: https://github.com/patternfly/pf-codemods/pull/386#pullrequestreview-1381949824)

jenny-s51 (commented):
Great points @thatblindgeye , thank you! I've updated this PR to add the `data-codemods=true` attribute to `Table` which is the method that was used in v4 to prevent any additional changes to components when running codemods multiple times

wise-king-sullyman (approved):

thatblindgeye:
@jenny-s51 So I tried having both this and the other table codemod locally, and running the --fix flag causes TableComposable to be renamed to Table, but also updates that new Table import/usage to the deprecated table despite the `data-codemod` attribute.

We could maybe try tweaking the moveSpecifier helper to check for that `data-codemods` attribute on a JSX Element, and if present don't update the import path for that specific import... We could also put one of these rules in the beta array so that it has to be explicitly ran with the --only flag, but that probably wouldn't be too great a solution. @gitdallas wdyt?

gitdallas:
Okay so trying to understand...

`import { TableComposable } from '@patternfly/react-table'; <TableComposable>Body</TableComposable>`

becomes

`import { Table } from '@patternfly/react-table'; <Table data-codemods="true">Body</Table>`

from this... then the other codemod turns into

`import { Table } from '@patternfly/react-table/deprecated'; <Table data-codemods="true">Body</Table>`

Yea, I think ignoring within the helper could make sense.

jenny-s51 (commented):
Thanks @thatblindgeye @gitdallas good catch - working on updating the helper but as you mentioned this is quite a complex case and requires non-trivial changes deep within helper logic ... I have a solution in development and will update once the changes have been implemented

jenny-s51 (commented):
Updated logic to ignore the rule if the data-codemods attribute exists - this should now run as expected however a race condition between codemods is preventing the table codemods updating the import statements correctly

As discussed in our codemods team meeting today, this PR and #330 are on hold while https://github.com/patternfly/pf-codemods/issues/426 is investigated

packages/eslint-plugin-pf-codemods/lib/rules/v5/table-rename-TableComposable.js
```diff
@@ -0,0 +1,96 @@
...
+  meta: {fixable: "code"},
+  create: function (context) {
+    const imports = getPackageImports(context, "@patternfly/react-table");
+    const tableComposableImport = imports.find(
+      (imp) => imp.imported.name === "TableComposable"
+    );
+
+    const tableImportsFiltered = imports.filter(
+      (imp) => imp.imported.name === "TableComposable"
+    );
```
thatblindgeye, line 14: Looks like we can remove this since it's not being used

thatblindgeye (changes requested):
Quick change below.

@gitdallas Jenny and I spoke about the issue still occurring where TableComposable is renamed to Table, but if the fix flag is ran again it'll update that to TableDeprecated despite having the codemods attribute. 

I had thrown together a tweak to the moveSpecifiers logic (here's the commit from my fork which is based off of this PR branch: https://github.com/thatblindgeye/pf-codemods/commit/79ae85c9f56f62b62b8c8f9b65e7a48bb688c52a). wdyt? It's a rather limited scope (things like interfaces or Context may not be affected, but right now it's really only this TableComposable scenario that is an issue), and it may depend on how likely the fix flag being ran multiple times is to occur.

gitdallas:
@thatblindgeye looks like it should work... be careful with the attr.name.name though 😆

packages/eslint-plugin-pf-codemods/index.js
```diff
@@ -29,10 +29,10 @@ const warningRules = [
 ];
 
 // rules that will run before other rules (move to deprecated?)
-const setupRules = [];
+const setupRules = ["table-update-deprecatedPath"];
 
 // rules that will run after other rules (cleanup imports?)
-const cleanupRules = [];
+const cleanupRules = ["table-rename-TableComposable"];
```
gitdallas, line 35: do we need to add this to the cleanup rules? i was thinking for this case we'd just need to add the deprecated stuff to the setup.

jenny-s51: Nice catch, removed

packages/eslint-plugin-pf-codemods/lib/helpers.js
```diff
@@ -10,11 +10,23 @@ function moveSpecifiers(
...
       fromPackage,
       importNames
     );
+    const allElements = getAllJSXElements(context);
+    const importSpecifiersToMove = fromPackageImports.filter((fromImport) => {
+      const foundElement = allElements.find(
+        (el) => el?.openingElement?.name?.name === fromImport.local?.name
+      );
+      const hasDataAttr = foundElement?.openingElement.attributes?.find(
+        (attr) => attr.name.name === "data-codemods"
```
thatblindgeye, line 24: Per Dallas' comment above that I forgot to update:

```suggestion
        (attr) => attr.name?.name === "data-codemods"
```

thatblindgeye (approved):

packages/pf-codemods/README.md
```diff
@@ -1901,8 +1901,6 @@ The following imports will be affected by this deprecation:
 
 #### Examples
 
-In:
```
gitdallas, line 1904: not a big deal but i think this was accidentally removed?

packages/eslint-plugin-pf-codemods/lib/rules/v5/table-rename-TableComposable.js
```diff
@@ -0,0 +1,92 @@
...
+                openingTagAttributes.filter(
+                  (attr) => attr.name?.name === "data-codemods"
+                ).length;
+              if (hasDataAttr) {
+                return;
+              }
+            }
+            if (
+              nodeName.includes("TableComposable") &&
+              importedNode.imported.name === importedNode.local.name // don't rename an aliased component
```
gitdallas, line 39: This would catch examples where we might have `import { Table as TableComposable }`?

I notice we are only looking for nodeName = "TableComposable" or "Table", so if someone is using an alias in general (`Table as PFTable` for example) we won't catch that. Is that by design? I know there's  other stuff happening with the setup rule and such so just want to make sure I understand well enough.

packages/eslint-plugin-pf-codemods/lib/rules/v5/table-rename-TableComposable.js
```diff
@@ -0,0 +1,92 @@
...
+              });
+            }
+          },
+          ImportSpecifier(node) {
+            const importedName = node.imported.name;
+            if (importedName === "TableComposable") {
+              const tableName = importedName.replace(
+                "TableComposable",
+                "Table"
+              );
```
gitdallas, line 76: wouldn't this be the same as:
```suggestion
              const tableName = "Table"
```

packages/eslint-plugin-pf-codemods/lib/rules/v5/table-rename-TableComposable.js
thatblindgeye: @jenny-s51 so I think I see what Dallas is talking about here. In the `test.tsx` file, if I import `TableComposable as TC`, the data-codemods attribute doesn't get applied on fix, so it'll eventually get its import path updated to deprecated as well. Using an alias, the `nodeName` will end up being TC so this if block will never be true.

From there, another issue is that the `imported name === local name` conditional is not applying the data attribute (as well as the `needsDataAttr` variable on line ~42). We'd want the data attribute applied even on an alias, but we could omit the rename if an alias exists.

jenny-s51: Thanks @gitdallas and @thatblindgeye for extra clarification - I've updated the logic here to account for cases when imports are aliased. If we are using an aliased import, the data attribute now gets applied on the aliased component, as Eric suggested. This prevents the import path from getting updated to deprecated

packages/eslint-plugin-pf-codemods/lib/rules/v5/table-rename-TableComposable.js
```diff
@@ -0,0 +1,91 @@
...
+              const hasDataAttr =
+                openingTagAttributes &&
+                openingTagAttributes.filter(
+                  (attr) => attr.name?.name === "data-codemods"
+                ).length;
+              if (hasDataAttr) {
+                return;
+              }
+            }
+            if (nodeName === "TableComposable" || isAliased) {
```
gitdallas, line 39: wouldn't this catch on any node that is using an alias import as long as a tableComposable exists?

i think we'd need to check that `nodeName === tableComposableImport.local.name`

packages/eslint-plugin-pf-codemods/lib/rules/v5/table-rename-TableComposable.js
```diff
@@ -0,0 +1,91 @@
...
+      : {
+          // update opening/closing elements
+          JSXIdentifier(node) {
+            const nodeName = node.name;
+            const importedNode = imports.find(
+              (imp) => imp.local.name === nodeName
+            );
+            const isAliased =
+              importedNode?.local?.name !== importedNode?.imported?.name;
+            if (nodeName === "Table") {
```
gitdallas, line 23: Would we want to check for an import with `imported.name === "Table"` and that it's `local.name === nodeName` here?

gitdallas (changes requested):

packages/eslint-plugin-pf-codemods/lib/rules/v5/table-rename-TableComposable.js
jenny-s51: Yes nice catch, thanks

gitdallas:
quite a few changes since reviews so gonna re-request

wise-king-sullyman (approved):

packages/eslint-plugin-pf-codemods/lib/rules/v5/table-rename-TableComposable.js
```diff
@@ -0,0 +1,97 @@
...
+                }
+                return tagName.replace("TableComposable", "Table");
+              };
+              const isOpeningTag =
+                context.getSourceCode().getTokenBefore(node).value !== "/";
+              const newOpeningParentTag = needsDataAttr
+                ? addDataAttr(updateTagName(node.parent))
+                : updateTagName(node.parent);
+              const newClosingTag = updateTagName(node);
+              context.report({
```
gitdallas, line 64: technically the message wouldn't be right if using an alias, although for opening tag we'd still want a report to add the data-attr.

for closing tag though we don't need to do anything, so might be worth doing
```suggestion
              (isOpeningTag || !isAliased) && context.report({
```

gitdallas (changes requested):

packages/eslint-plugin-pf-codemods/lib/rules/v5/table-rename-TableComposable.js
```diff
@@ -0,0 +1,97 @@
...
+              }
+            }
+            if (
+              nodeName === "TableComposable" ||
+              nodeName !== tableComposableImport.local.name
+            ) {
+              const tableName = nodeName.replace("TableComposable", "Table");
+              const needsDataAttr = tableName === "Table" || isAliased;
+              const addDataAttr = (jsxStr) =>
+                `${jsxStr.slice(0, -1)} data-codemods="true">`;
```
gitdallas, line 49: should we also consider self closing tags? this wouldn't work for those.. perhaps
```suggestion
                `${jsxStr.replace(/(\/)*>$/, ' data-codemods="true" $1>`;
```

thatblindgeye (approved):
Other than the above and needing a rebase, looks good.

jenny-s51 (commented):
Thank you @gitdallas for your thoughtful and helpful review comments - updated based on your suggestions

jenny-s51 (commented):
Thank you @gitdallas for your thoughtful and helpful review comments - updated based on your suggestions

gitdallas (approved):

packages/eslint-plugin-pf-codemods/lib/rules/v5/table-rename-TableComposable.js
```diff
@@ -0,0 +1,86 @@
...
+              openingTagAttributes &&
+              openingTagAttributes.filter(
+                (attr) => attr.name?.name === "data-codemods"
+              ).length;
+            const isAliased =
+              importedNode?.local?.name !== importedNode?.imported?.name;
+            if (importedNode?.imported?.name === "Table" && hasDataAttr) {
+              // if data-codemods attribute, do nothing
+              return;
+            } else if (importedNode?.imported?.name === "TableComposable") {
```
gitdallas, line 36: Table && hasDataAttr doesn't seem to be necessary as we aren't going to do anything if `imported.name === 'Table'` anyway (only if "TableComposable").

I think we can just remove that check and the `hasDataAttr` with it.

this codemod will only affect TableComposable which will never have a codemod data attr as we always convert to table at the same time (so that if they run the codemods again it doesn't try to deprecate it after converting it)

gitdallas (approved):
nit about some unnecessary code

packages/eslint-plugin-pf-codemods/lib/rules/v5/table-rename-TableComposable.js
```diff
@@ -0,0 +1,83 @@
...
+            const parentNode = node.parent;
+            const isOpeningTag = parentNode.type === "JSXOpeningElement";
+            const openingTagAttributes = isOpeningTag
+              ? parentNode.attributes
+              : parentNode.parent.openingElement?.attributes;
+            const hasDataAttr =
+              openingTagAttributes &&
+              openingTagAttributes.filter(
+                (attr) => attr.name?.name === "data-codemods"
+              ).length;
```
gitdallas, line 30: unused var
```suggestion
```

gitdallas (changes requested):
unused var below, also a conflict popped up

gitdallas (approved):

PR #389, opened 2023-04-10
feat(misc icons): warn remove size, color, and noVerticalAlign props

### LINKED ISSUE #370: Misc icons - removed size, noVerticalAlign, and color props
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/5275

_Required actions:_
1. Build codemod
2. Build test

jenny-s51:
Closes #370


packages/eslint-plugin-pf-codemods/test/rules/v5/react-icons-remove-props.js
```diff
@@ -0,0 +1,34 @@
...
+const rule = require("../../../lib/rules/v5/react-icons-remove-props");
+
+ruleTester.run("react-icons-remove-props", rule, {
+  valid: [
+    {
+      code: `import { FrogIcon } from '@patternfly/react-icons'; <FrogIcon  />`,
+    },
+    {
+      // No @patternfly/react-core import
+      code: `<FrogIcon  />`,
```
thatblindgeye, line 11:
```suggestion
      // No @patternfly/react-icons import
      code: `<FrogIcon size color noVerticalAlign  />`,
```

Just to test that these props won't throw an error for a non PF icons import

thatblindgeye (changes requested):

thatblindgeye (approved):

wise-king-sullyman:
Could we make this a bit smarter? Going through this in topology right now and if it could automatically add the width/height props that would've saved me a bunch of time debugging.

gitdallas:
@wise-king-sullyman  - just to clarify what we might want to do:

https://github.com/patternfly/patternfly-react/pull/5275/files#diff-e1db0b5cfe1524529c167e79fc188fbe2f5138a4f6e60847a7025832fcbb6556L32

`size="lg"` => `width="2em" height="2em"`

etc based on 

https://github.com/patternfly/patternfly-react/pull/5275/files#diff-3c2447197cf79209a86df8da5c9cbddeff9e652e7a68fda27803104a25298474L10

    case IconSize.sm:
      return '1em';
    case IconSize.md:
      return '1.5em';
    case IconSize.lg:
      return '2em';
    case IconSize.xl:
      return '3em';
    default:
      return '1em';

sound right?

wise-king-sullyman:
The issue I was having in topology actually ended up being something else 😆 

But yeah @gitdallas that's what I had in mind, and I do still feel like that would be a big improvement for consumers.

jenny-s51 (commented):
There was some good discussion with @wise-king-sullyman and @mcoker regarding how we want to address this breaking change for consumers.

Per @mcoker:
> There isn't a 1:1 match between the old react-icons size prop values and the sizes we want icons and things in PF. I think that would be a good opportunity for any application to review their use of color and size and see if they can use what PF provides via something like the <Icon> component. Meaning use one of our supported size and color values for <Icon> instead of something custom

Therefore we don't want to delete these props for the consumer because there isn't a 1:1 mapping for whatever custom settings the consumer had, and the size/color options PF supports within the `Icon` component. 

This codemod now warns and suggests for the consumer to add custom styling by wrapping their react-icon with the PF `Icon` component.

packages/eslint-plugin-pf-codemods/lib/rules/v5/react-icons-warn-remove-props.js
```diff
@@ -0,0 +1,34 @@
...
+    const imports = getPackageImports(context, "@patternfly/react-icons");
+
+    const iconImport = imports.filter((specifier) =>
+      specifier.imported.name.includes("Icon")
+    );
+    return imports.length === 0
+      ? {}
+      : {
+          JSXOpeningElement(node) {
+            if (
```
gitdallas, line 16: let's also check that `node.name.name === iconImport.local.name`, otherwise this could match on anything that has a size/color/noVerticalAlign attribute as long as there is a pf icon import.

packages/eslint-plugin-pf-codemods/lib/rules/v5/react-icons-warn-remove-props.js
```diff
@@ -0,0 +1,34 @@
+const {getPackageImports} = require("../../helpers");
+
+// https://github.com/patternfly/patternfly-react/pull/5275
+module.exports = {
+  meta: {fixable: "code"},
+  create: function (context) {
+    const imports = getPackageImports(context, "@patternfly/react-icons");
+
+    const iconImport = imports.filter((specifier) =>
```
gitdallas, line 9: could do `find` instead

packages/eslint-plugin-pf-codemods/lib/rules/v5/react-icons-warn-remove-props.js
```diff
@@ -0,0 +1,34 @@
...
+      specifier.imported.name.includes("Icon")
+    );
+    return imports.length === 0
```
gitdallas, line 12: probably want to check on `iconImport` here

gitdallas (changes requested):

jenny-s51 (commented):
Thanks @gitdallas - great points. Updated

wise-king-sullyman:
Should the reports thrown by this rule be errors or warnings?

wise-king-sullyman (approved):

packages/eslint-plugin-pf-codemods/lib/rules/v5/react-icons-warn-remove-props.js
```diff
@@ -0,0 +1,36 @@
...
+    const iconImport = imports.find((specifier) =>
+      specifier.imported.name.includes("Icon")
+    );
+
+    return !iconImport
+      ? {}
+      : {
+          JSXOpeningElement(node) {
+            if (
+              iconImport &&
```
gitdallas, line 18: nit.. not needed as we are within the else of !iconImport

packages/eslint-plugin-pf-codemods/index.js
```diff
@@ -23,6 +23,7 @@ const warningRules = [
   "overflowMenu-warn-updated-dropdownItem",
   "popover-warn-appendTo-default",
   "react-dropzone-warn-upgrade",
+  "react-icons-warn-remove-props",
```
gitdallas, line 26: i think to @wise-king-sullyman 's point, this should probably be an error. these properties were removed

gitdallas (changes requested):

packages/eslint-plugin-pf-codemods/index.js
gitdallas: oh... was i wrong? would this not be an error in the code?

jenny-s51: @gitdallas I think either would make sense here - should it be a warning since we don't specifically provide the fix? Or can it be an error even if we don't provide the fix

gitdallas (approved):

wise-king-sullyman (approved):

PR #394, opened 2023-04-12
Absolute paths react core

KKoukiou:


packages/eslint-plugin-pf-codemods/lib/helpers.js
```diff
@@ -186,7 +186,7 @@ function getPackageImports(context, packageName, importNames = []) {
   const specifiers = context
     .getSourceCode()
     .ast.body.filter((node) => node.type === "ImportDeclaration")
-    .filter((node) => node.source.value === packageName)
+    .filter((node) => node.source.value.startsWith(packageName))
```
gitdallas, line 189: We can't just have starts with, most cases are using `@patternfly/react-core` which would include `/deprecated` and `/next` and such that we might not want to include with each rule that uses this.

gitdallas (changes requested):

PR #396, opened 2023-04-13
chore(Td): rename disable prop in TdSelectType and TdActionsType

adamviktora:
Closes https://github.com/patternfly/pf-codemods/issues/347 and closes https://github.com/patternfly/pf-codemods/issues/354

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/td-TdSelectType-TdActionsType-rename-disable.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/td-TdSelectType-TdActionsType-rename-disable.js
@@ -25,11 +25,6 @@ module.exports = {
                 ["select", "actions"].includes(attr.name?.name)
             );
 
-            const interfaceMap = {
-              select: "TdSelectType",
-              actions: "TdActionsType",
-            };
-
             const getObjectProp = (expr, propName) =>
               expr.properties.find(
                 (prop) =>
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin-pf-codemods/lib/rules/v5/td-TdSelectType-TdActionsType-rename-disable.js
```diff
@@ -0,0 +1,94 @@
...
+
+    return !tableTdImport
+      ? {}
+      : {
+          JSXOpeningElement(node) {
+            if (tdName !== node.name?.name) {
+              return;
+            }
+
+            const attributes = node.attributes.filter(
```
thatblindgeye, line 23: Nit: var name could be `attribuesToUpdate` or something like that, since we're only looking for specific attributes vs all of them

jenny-s51 (approved):
Nice work @adamviktora ! LGTM

thatblindgeye (approved):

packages/eslint-plugin-pf-codemods/lib/rules/v5/td-TdSelectType-TdActionsType-rename-disable.js
```diff
@@ -0,0 +1,94 @@
...
+                doReport(getObjectProp(expr, "disable"));
+              }
+
+              if (expr.type === "Identifier") {
+                console.log(expr.loc);
+
+                let scope = context.getScope();
+                while (scope !== null) {
+                  const variable = scope.variables.find(
+                    (v) => v.name === expr.name
```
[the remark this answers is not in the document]
gitdallas: @thatblindgeye 
just fyi `var` would have caused an error (keyword in js). i also tend to use short names like this for one-liners though, and would actually prefer over variable shadowing the `variable` variable 😄 

but no biggie

packages/eslint-plugin-pf-codemods/lib/rules/v5/td-TdSelectType-TdActionsType-rename-disable.js
```diff
@@ -0,0 +1,92 @@
...
+                  const variable = scope.variables.find(
+                    (variable) => variable.name === expr.name
+                  );
+                  if (!variable) {
+                    scope = scope.upper;
+                    continue;
+                  }
+                  if (
+                    variable.references.some(
+                      (ref) => ref.identifier.loc === expr.loc
```
gitdallas, line 76: double checking that the variable found is the correct one?

since we are looking up the scope tree by name, have you identified a scenario that this would not be the same?

packages/eslint-plugin-pf-codemods/lib/rules/v5/td-TdSelectType-TdActionsType-rename-disable.js
```diff
@@ -0,0 +1,92 @@
...
+                    )
+                  ) {
+                    variable.defs.forEach((def) => {
```
gitdallas, line 79: This is more me trying to understand the structure of the variable interface... but would there ever be more than a single definition for a variable?

btw, I tried doing 

      import { Td } from "@patternfly/react-table";
      const isDisabled = true;
      let myObj = {"disable": isDisabled};
      myObj = { "disable": false};
      <Td select={myObj} actions={{disable: false}} ></Td>`,

and it converted to

            import { Td } from "@patternfly/react-table";
            const isDisabled = true;
            let myObj = {isDisabled: isDisabled};
            myObj = { "disable": false};
             <Td select={myObj} actions={{isDisabled: false}} ></Td>

note that it correctly updated the initialization but not the update, this is not a breaker for me as it's probably an edge case that nobody would be using.. but if you wanted to look into it

gitdallas (approved):
great job, i have some nits/questions but would be fine merging

packages/eslint-plugin-pf-codemods/lib/rules/v5/td-TdSelectType-TdActionsType-rename-disable.js
adamviktora: About the _would there ever be more than a single definition for a variable?_ It may be a case if we would ignore scopes, but each variable has its scope, so I am also confused, why is the `Variable.defs` property an array.

gitdallas (approved):
nice work

wise-king-sullyman (approved):

PR #398, opened 2023-04-13
chore(BackgroundImage/LoginPage): updated props

### LINKED ISSUE #392: BackgroundImage/LogIn - removed/updated props
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/8931

- The `filter` prop was removed from BackgroundImage and `backgroundImgAlt` from LogIn
- the `src` prop type was updated to just a string for BackgroundImage (similar update for LogIn's `backgroundImgSrc`

Should provide a fixer for removing `filter` and `backgroundImgAlt`, and possibly just a non-fixing error for the `src`/`backgroundImgSrc` prop since we may not know what property from the map a consumer would want to use.

_Required actions:_
1. Build codemod
2. Build test

thatblindgeye:
Closes #392

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/backgroundImageSrcMap-remove-interface.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/backgroundImageSrcMap-remove-interface.js
@@ -26,13 +26,15 @@ module.exports = {
                 fix(fixer) {
                   const tokenAfter = context
                     .getSourceCode()
-                    .getTokenAfter(backgroundImgSrcMapSpecifier).value;
+                    .getTokenAfter(backgroundImgSrcMapSpecifier);
 
                   const { range } = backgroundImgSrcMapSpecifier;
                   return node.specifiers.length > 1
                     ? fixer.removeRange([
                         range[0],
-                        tokenAfter === "," ? range[1] + 1 : range[1],
+                        tokenAfter.value === ","
+                          ? tokenAfter.range[1]
+                          : range[1],
                       ])
                     : fixer.remove(node);
                 },
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/backgroundImageSrcMap-remove-interface.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/backgroundImageSrcMap-remove-interface.js
@@ -1,0 +1,63 @@
+const { getPackageImports } = require("../../helpers");
+
+// https://github.com/patternfly/patternfly-react/pull/8931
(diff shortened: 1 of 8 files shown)


wise-king-sullyman (approved):

packages/eslint-plugin-pf-codemods/lib/rules/v5/backgroundImageSrcMap-remove-interface.js
```diff
@@ -0,0 +1,63 @@
...
+                fix(fixer) {
+                  const tokenAfter = context
+                    .getSourceCode()
+                    .getTokenAfter(backgroundImgSrcMapSpecifier).value;
+
+                  const { range } = backgroundImgSrcMapSpecifier;
+                  return node.specifiers.length > 1
+                    ? fixer.removeRange([
+                        range[0],
+                        tokenAfter === "," ? range[1] + 1 : range[1],
```
gitdallas, line 35: this will most likely be fine (people probably won't do `import { foo , bar }` with the space)... but we could just the token after rather than the value of it, check the value of it.. then use takenAfter.range[1] if it's a comma

packages/eslint-plugin-pf-codemods/lib/rules/v5/backgroundImageSrcMap-remove-interface.js
```diff
@@ -0,0 +1,63 @@
...
+              typeAnnotationProperty.name !==
+                backgroundImgSrcMapImport.local.name
+            ) {
+              return;
+            }
+            context.report({
+              node,
+              message: `The BackgroundImageSrcMap interface has been removed.`,
+              fix(fixer) {
+                return fixer.remove(node.typeAnnotation);
```
gitdallas, line 57: nice

packages/eslint-plugin-pf-codemods/lib/rules/v5/loginPage-update-props.js
```diff
@@ -0,0 +1,57 @@
...
+            }
+
+            renameProps0(context, loginPageImports, node, {
+              LoginPage: {
+                backgroundImgAlt: "",
+              },
+            });
+
+            const backgroundImgSrcProp = node.attributes.find(
+              (attribute) => attribute.name.name === "backgroundImgSrc"
```
gitdallas, line 31:
```suggestion
              (attribute) => attribute.name?.name === "backgroundImgSrc"
```

gitdallas (changes requested):
good, little things

gitdallas (approved):

PR #405, opened 2023-04-17
feat: No unused imports rule

### LINKED ISSUE #397: Create v5 version of no-unused-import codemod
Based on conversation during a Meet, we want to create a v5 version of the `no-unused-imports` codemod that v4 has. Could be a simply copy+paste job, or we can see if there's any refactoring we can do.

We also should add the rule to the new `betaRuleNames` array in the `eslint-plugin-pf-codemods/index.js` file, as we only want this codemod ran when the `--only` flag is applied (basically a "run this codemod only when you are done running/fixing all other codemods).

_Required actions:_
1. Build codemod
2. Build test

adamviktora:
Closes #397

Did some refactoring from the v4 rule:
- removing imports does not keep trailing commas, which were causing the file to by syntactically incorrect
- tokens are filtered with `JSXIdentifier` and `Identifier` types, so there are no misleadings, e.g. in this case: `<Title>Button</Title>` if there was an unused `Button` import, it was not removed in the previous version
- errors and fixes are more compact, 1 error and 1 group of fixes per package
- if there are no used imports from the package left, the whole import declaration is removed

The thing I was not been able to repair is to use warnings, instead of errors. There is a clash with the beta rules, which are not updated to use warnings / errors.

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/no-unused-imports-v5.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/no-unused-imports-v5.js
@@ -1,0 +1,70 @@
+// Cleanup from other rules
+module.exports = {
+  meta: { fixable: "code" },
+  create: function (context) {
+    const sourceCode = context.getSourceCode();
+
+    const allTokens = [
+      ...new Set(
+        sourceCode.ast.body
+          .filter((node) => node.type !== "ImportDeclaration")
+          .map((node) =>
+            sourceCode
+              .getTokens(node)
+              .filter((token) =>
+                ["JSXIdentifier", "Identifier"].includes(token.type)
+              )
+              .map((token) => token.value)
+          )
+          .flat()
+      ),
+    ];
+
+    return {
+      ImportDeclaration(node) {
+        if (!/@patternfly\/react/.test(node.source.value)) {
+          return;
+        }
+
+        const unusedImports = node.specifiers.filter(
+          (spec) => !allTokens.includes(spec.local.name)
+        );
+
+        if (unusedImports.length === 0) {
+          return;
+        }
+
+        context.report({
+          node,
+          message: `unused PatternFly import${
+            unusedImports.length > 1 ? "s" : ""
+          } ${unusedImports.map((spec) => spec.local.name).join(", ")} from '${
+            node.source.value
+          }'`,
+          fix(fixer) {
+            const getEndRange = (spec) => {
+              const nextComma = sourceCode.getTokenAfter(spec);
+              return nextComma.value === ","
+                ? sourceCode.getTokenAfter(nextComma).range[0]
+                : spec.range[1];
+            };
+
+            const removeWholeImport = () => {
+              const tokenAfter = sourceCode.getTokenAfter(node);
+              return [
+                fixer.remove(node),
+                ...(tokenAfter === ";" ? [fixer.remove(tokenAfter)] : []),
+              ];
+            };
+
+            return unusedImports.length === node.specifiers.length
+              ? removeWholeImport()
+              : unusedImports.map((spec) =>
+                  fixer.removeRange([spec.range[0], getEndRange(spec)])
+                );
+          },
+        });
+      },
+    };
+  },
+};
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/no-unused-imports-v5.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/no-unused-imports-v5.js
@@ -43,22 +43,14 @@ module.exports = {
           }'`,
           fix(fixer) {
             const getEndRange = (spec) => {
-              const nextComma = sourceCode.getTokenAfter(spec);
-              return nextComma.value === ","
-                ? sourceCode.getTokenAfter(nextComma).range[0]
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin-pf-codemods/lib/rules/v5/no-unused-imports-v5.js
```diff
@@ -0,0 +1,70 @@
...
+          } ${unusedImports.map((spec) => spec.local.name).join(", ")} from '${
+            node.source.value
+          }'`,
+          fix(fixer) {
+            const getEndRange = (spec) => {
+              const nextComma = sourceCode.getTokenAfter(spec);
+              return nextComma.value === ","
+                ? sourceCode.getTokenAfter(nextComma).range[0]
+                : spec.range[1];
+            };
```
thatblindgeye, line 50: Nit: `tokenAfter` instead of `nextComma`, since the token won't always be a comma

thatblindgeye (changes requested):

packages/eslint-plugin-pf-codemods/lib/rules/v5/no-unused-imports-v5.js
```diff
@@ -0,0 +1,70 @@
...
+
+            const removeWholeImport = () => {
+              const tokenAfter = sourceCode.getTokenAfter(node);
+              return [
+                fixer.remove(node),
+                ...(tokenAfter === ";" ? [fixer.remove(tokenAfter)] : []),
+              ];
+            };
```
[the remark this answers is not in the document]
adamviktora: Yeah, good idea, I still had in mind removing the comma after specifier (which is necessary), but the semicolon seems to be the part of the node.

gitdallas (commented):
If you'd like to not have leftover commas, here is code from one of my closed PRs awhile back that you might be able to use


	...specifiersToRemove.map( node => {
                const before = context.getTokenBefore(node);
                // checking against local as coming from the right, alias will show up first if exists and we want to recognize that
                const isImportBeforeFixed = specifiersToRemove.map(n => n.local?.name).includes(context.getTokenBefore(before)?.value);
                const after = context.getTokenAfter(node);
                let isRemainingImportsFixed = true;
                for(let n = after; n.value !== "}"; n = context.getTokenAfter(n)) {
                  if( n.type !== "Punctuator" && !specifiersToRemoveImportNameList.includes(n.value) ) {
                    isRemainingImportsFixed = false;
                    break;
                  }
                }
                let range = node.range;
                if (after.value === ',') {
                  range[1] = context.getTokenAfter(after).range[0];
                } 
                if (isRemainingImportsFixed && before.value === ',' && !isImportBeforeFixed) {
                  range[0] = context.getTokenBefore(before).range[1];
                }
                return fixer.replaceTextRange(range, '');
              })

adamviktora:
thanks for the comma-removing code, there is no problem with removing commas though, it should be handled by this code:

`          fix(fixer) {
            const getEndRange = (spec) => {
              const tokenAfter = sourceCode.getTokenAfter(spec);
              return tokenAfter.value === ","
                ? sourceCode.getTokenAfter(tokenAfter).range[0]
                : spec.range[1];
            };

            return unusedImports.length === node.specifiers.length
              ? fixer.remove(node)
              : unusedImports.map((spec) =>
                  fixer.removeRange([spec.range[0], getEndRange(spec)])
                );
          },`

gitdallas:
Just make sure that it would work with like... we are removing Foo, Bar, and Baz

tests:

- Foo, Bar, Baz
- Table, Foo, Button, Bar, OptionList, Baz

i think with the code above, `Foo, Bar, Baz` might run into "overlap" issues (feel like i've got that in the past trying something like that).

gitdallas:
@thatblindgeye @thatblindgeye @jenny-s51 - what do you guys think about taking this out of `beta` list and putting it into `cleanup`? any reason not to?

jenny-s51 (commented):
@gitdallas I agree this would make sense as a cleanup rule

thatblindgeye:
@gitdallas Seems like it works fine when testing with the empty state fixes (since that had given us issues when trying to apply fixes).

adamviktora:
> Just make sure that it would work with like... we are removing Foo, Bar, and Baz
> 
> tests:
> 
>     * Foo, Bar, Baz
> 
>     * Table, Foo, Button, Bar, OptionList, Baz
> 
> 
> i think with the code above, `Foo, Bar, Baz` might run into "overlap" issues (feel like i've got that in the past trying something like that).

Tried these and other cases, works fine - when all imports from the current source package are unused, it removes the whole import.

packages/pf-codemods/README.md
```diff
@@ -1405,6 +1405,38 @@ We've update the `aria-label` prop on MenuItemAction, making it required instead
 
 The placement Nav flyouts in the DOM has been changed, if you have Nav elements with flyouts you may need to update some selectors or snapshots in your test suites. This rule will raise a warning, but will not make any changes.
 
+### no-unused-imports-v5
+
+This rule removes all unused imports from `patternfly/react` packages. It is a special rule which will not run by default. It should be run after all the other rules as a final cleanup using the option `--only no-unused-imports-v5 --fix`.
```
gitdallas, line 1410: will need to update this now that it's in cleanup rather than beta

gitdallas (changes requested):

PR #414, opened 2023-04-20
feat(general): handle esm imports

### LINKED ISSUE #282: handle absolute paths on imports (not a codemod)
It's possible that a consumer could import like:

```
import { Select } from '@patternfly/react-core/dist/esm/components/Select/Select';
import { SelectOptionObject } from '@patternfly/react-core/dist/esm/components/Select/SelectOption';
```

which most of our codemods would not pick up.

only possible issue that i can think of with just checking "starts with @patternfly/{package}" might be that we need could have the same specifier from next... so maybe check for (or against) that?

gitdallas:
closes #282

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/helpers.js
+++ b/packages/eslint-plugin-pf-codemods/lib/helpers.js
@@ -29,6 +29,7 @@ function moveSpecifiers(
       if (toParts[0] === "@patternfly" && toParts.length === 3) {
         fromParts.splice(4, 0, toParts[2])
         toPackage = fromParts.join("/");
+        console.log("TOOOPACKAGEEEEEE\nXXXX", toPackage);
       }
     }
 
--- a/packages/eslint-plugin-pf-codemods/lib/helpers.js
+++ b/packages/eslint-plugin-pf-codemods/lib/helpers.js
@@ -177,6 +177,17 @@ function moveSpecifiers(
   };
 }
 
+function pfPackageMatches(packageName, nodeSrc) {
+  const parts = packageName.split("/");
+  const regex = new RegExp('^' +
+    parts[0] + '\/' + parts[1] +
+    '(\/dist\/esm)?' +
+    (parts[2] ? ('\/' + parts[2]) : '') +
+    '(\/components\/.*)?$'
+  );
+  return regex.test(nodeSrc)
+}
+
 /**
  *
  * @param context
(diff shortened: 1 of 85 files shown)


packages/eslint-plugin-pf-codemods/lib/helpers.js
```diff
@@ -186,7 +211,12 @@ function getPackageImports(context, packageName, importNames = []) {
   const specifiers = context
     .getSourceCode()
     .ast.body.filter((node) => node.type === "ImportDeclaration")
-    .filter((node) => node.source.value === packageName)
+    .filter((node) => {
+      if(packageName.startsWith("@patternfly")) {
+        return pfPackageMatches(packageName, node.source.value);
+      }
+      return node.source.value === packageName
+    })
```
thatblindgeye, line 219: Do we need this `if` block to return one of two things here considering other helpers just use `pfPackageMatches` without any sort of `startsWith` check?

thatblindgeye (commented):
File comment below. Also, should we add tests to the `charts-...` test files? The `charts-lightThemeObjects` file has a test I threw in to take into account importing from `/another/path`, though that was based on Dan mentioning something being imported from `@patternfly/react-charts/dist/js/...` (rather than `/esm`.

packages/eslint-plugin-pf-codemods/lib/helpers.js
thatblindgeye: I wouldn't block over making it a `packageMatches` vs `pfPackageMatches` helper. We can always cross that bridge when we come to it if anything, but could make sense to try and open it up beyond just patternfly now.

Any comment re: the `charts-` files?

gitdallas: @thatblindgeye  - just pushed with a couple `charts` tests. also updated the matcher to check `dist/esm` or `dist/js`

gitdallas: as far as opening beyond just pf... i guess i'm not familiar enough with this to know that other packages would have similar structure with esm/js exports, @thatblindgeye

thatblindgeye (approved):
Would be a nice-to-have adding tests to the `-remove-ChartThemeVariant`, `-remove-darkThemeObjects`, `-remove-themeVariant`, and `-reiszeObserver-import` chart tests, but that can be a followup if we'd want to look into what those import paths might actually look like

kmcfaul (approved):
🎉

wise-king-sullyman (approved):

PR #416, opened 2023-04-25
fix error calling getScope on source code instead of context

gitdallas:
now sure how this passed ci before, but it failed test/test:single for austin, eric, and myself


adamviktora:
according to documentation: https://eslint.org/docs/latest/extend/custom-rules#the-context-object
calling `getScope()` directly on the context object is deprecated and `SourceCode#getScope(node)` should be used instead
Screenshot 2023-04-25 at 15 43 22

what exactly is failing? yarn test:v5:single works fine for me (besides this error, but it is unrelated)
Screenshot 2023-04-25 at 15 49 38

adamviktora:
could it be related to node version possibly? i am using v18.15.0

Issue #417, opened 2023-04-25
FileUploadField - remove from beta rules list


wise-king-sullyman:
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/issues/8882

The code mod was already made, but we need to:

_Required actions:_
1. Remove from the beta rules list
2. Update rule to also change FileUpload's `onTextChange` as was done in 8882
3. Update readme to not state that the rule isn't in pf-react

PR #420, opened 2023-04-25
feat(FileUpload): Added events params to callbacks

### LINKED ISSUE #402: FileUpload- event param added to several callbacks
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/8960

_Required actions:_
1. Build codemod
2. Build test

wise-king-sullyman:
Closes #402

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/fileUpload-event-params-added-to-callbacks.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/fileUpload-event-params-added-to-callbacks.js
@@ -1,0 +1,12 @@
+const { addCallbackParam } = require("../../helpers");
+
+// https://github.com/patternfly/patternfly-react/pull/8960
+module.exports = {
+  meta: { fixable: "code" },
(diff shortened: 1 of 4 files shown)


thatblindgeye (commented):
Overall this looks good, just a couple questions/comments:

- Probably want to add this rule to the betaRule array (I believe that's what we landed on in terms of new codemods going in?)
- Though it wasn't part of the PR being linked currently, we could update onTextChange as part of this. We do already have a codemod for updating onTextChange on FileUpload*Field*, but not FileUpload. We could also update `fileUploadField-cb-param-updates` so that FileUpload is included in the component array there instead (possibly as part of any followup to that particular codemod, like #417 )

wise-king-sullyman:
Rather than add it to the betaRule array and have to come back and remove it later I just threw the do not merge tag on this until the PR is merged, is that adequate?

And I have a note in #417 to add the FileUpload onTextChange when I do that followup since I did that change under the FileUploadField PR.

thatblindgeye (approved):

gitdallas (approved):

PR #424, opened 2023-04-27
chore(helpers): renameProps helper refactoring

adamviktora:
- getting rid of the `renameProp` helper completely
- `renameProps` now accepts more parameters for each prop to modify
    - `newName`: new prop name
    - `message`: custom message, can be a string or a function accepting `node` - this is good for printing out the local node name, when using aliased imports
    - `replace`: indicates, that the whole prop should be replaced. There were lots of "not so correct" usages of the renameProps helper, where the whole prop has been replaced (e.g. `withToolBar` -> `categoryName="pf-codemod-category"`), but it previously worked in a way, that only the prop name was changed, so if there would be some code like `withToolBar={true}`, it would lead to `categoryName="pf-codemod-category"={true}`, that is why the replace prop is introduced
    - prop can still accept just a string as a new prop name, to keep backwards compatibility and keeping the simple usage, when there is no need for custom `message` or `replace` option
- whole structure of the `renames` object passed to `renameProps` is described in a docstring above the function

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/helpers.js
+++ b/packages/eslint-plugin-pf-codemods/lib/helpers.js
@@ -314,6 +314,34 @@ function renamePropsOnNode(context, imports, node, renames) {
   }
 }
 
+/**
+ * 
+ * @param {*} renames 
+ * structure of the renames object:
+ * {
+ *    ComponentOne: {
+        propA: {
+          newName: "newPropA",
+          message: (node) => `propA prop has been renamed to newPropA for ${node.name.name}, some custom message`, // message is optional, default message will always be provided
+        },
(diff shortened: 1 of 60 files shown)


gitdallas (approved):
lgtm, i think on the rename object i would have preferred a structure like...

```
[
  {
    component: "ComponentName",
    props: [ ... ]
  },
   ...
]
```

just because i prefer to stay away from basically variables as property names and having to do Object.keys and such... but this is closer to the original and it would be a pretty big effort to make changes now so i'm not going to worry about that lol

thatblindgeye (approved):

wise-king-sullyman (approved):

PR #440, opened 2023-05-04
feat(Select): deprecate select - update imports

### LINKED ISSUE #324: Select - update path to deprecated directory
Follow up to breaking change issue https://github.com/patternfly/patternfly-react/issues/8073

_Required actions:_
1. Build codemod
2. Build test

adamviktora:
Closes #324

I am not sure whether to apply the `moveSpecifiers` helper on all exports, or if components are sufficient. Anyway I applied it on all exports from the `deprecated/components/Select` folder including constants, enums and interfaces.


gitdallas (approved):

thatblindgeye (approved):
@gitdallas After the issue we came across with the Table PRs, I'm wondering if it would make sense to add a codemod that updates paths for Dropdown and Select from `/next` to just `react-core`. That'd just involve adding this codemod to that setup array, and making sure the data-codemods attribute is added to the Next Select/Dropdown. wdyt?

gitdallas:
@thatblindgeye  - I don't think we have any mods currently that would be impacted negatively... looks like Select changes are `onToggle` param changes and removing `removeFindDomNode`

Don't think either of these would hurt to work on both deprecated/next.. but we do want to make sure that they also affect the `deprecated` imports.

That being said... shouldn't hurt to add it to the setup and might help if we have some future mod on it?

PR #441, opened 2023-05-04
feat(PageHeader): deprecate PageHeader - update imports

### LINKED ISSUE #329: Page header - update path to deprecated directory
Follow up to breaking change issue https://github.com/patternfly/patternfly-react/issues/8602

_Required actions:_
1. Build codemod
2. Build test

adamviktora:
Closes #329

### DIFF
--- a/packages/pf-codemods/README.md
+++ b/packages/pf-codemods/README.md
@@ -1692,26 +1692,6 @@ import {
 </>;
 ```
 
-### pageheader-update-logoComponent [(#8655)](https://github.com/patternfly/patternfly-react/pull/8655)
-
-We've updated `PageHeader`'s logo to only be an anchor if an `href` is specified, otherwise it will be a `span`. Explicitly declared `logoComponent` properties will remain unchanged, but if it is not specified a default will be added.
-
-#### Examples
-
-In:
-
-```jsx
-<PageHeader />
-<PageHeader logoComponent="div" />
-```
-
-Out:
-
-```jsx
-<PageHeader logoComponent="a" />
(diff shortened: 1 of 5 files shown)


packages/pf-codemods/README.md
```diff
@@ -1519,6 +1519,52 @@ Out:
 <OverflowMenuDropdownItem itemId={0}>
 ```
 
+### pageHeader-deprecated [(#8854)](https://github.com/patternfly/patternfly-react/pull/8854)
+
+We've deprecated the `PageHeader` components. A fix will update code to point to the new deprecated import, but we recommend replacing with `Masthead` and its related components.
```
thatblindgeye, line 1524: Just to name them all in the rule description:

```suggestion
We've deprecated the `PageHeader`, `PageHeaderTools`, `PageHeaderToolsGroup`, and `PageHeaderToolsItem` components. A fix will update code to point to the new deprecated import, but we recommend replacing with `Masthead` and its related components.
```

gitdallas (approved):

gitdallas:
@thatblindgeye  - you think this and all import updates should be moved in setup? kind of makes it so we **should** know what we are dealing with when we get to the rest of the mods. should go back and add other import moves to setup? can we see any downsides? @wise-king-sullyman

thatblindgeye:
@gitdallas it might make sense to just put all import path updates in setup yeah. If we end up in a situation like the Table stuff again, it'd probably make it easier to write a codemod to avoid any "renamed components still get their import path updated" issues.

PR #452, opened 2023-05-09
fix(EmptyState): updated codemod for renaming actions

### LINKED ISSUE #369: EmptyState - refine component renaming fixer
Follow up to a [suggestion for possible improvement](https://github.com/patternfly/pf-codemods/pull/346#discussion_r1153706283)

Currently the fixer for the component renaming requires multiple runs to complete the fix, ideally this would be doable in a single run.

_Required actions:_
1. Refine fixer in emptystate-rename-components
2. Update tests to reflect that multiple passes are no longer needed

thatblindgeye:
Closes #369 

Simply removes some logic from the codemod to prevent the need for tests having to make multiple passes. The codemod itself won't remove unused imports, but since we have the new rule that runs on cleanup that should take care of things.


wise-king-sullyman (approved):

gitdallas (approved):
thanks for finding this

edit: i guess the story was already there, but thanks for connecting the dots :P

PR #459, opened 2023-05-09
fix(Nav): added event param as first param

### LINKED ISSUE #430: Navigation - update param order
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/8997

May also want to try adding a warning about `event` being removed from the `xItem` param object.

_Required actions:_
1. Build codemod
2. Build test

thatblindgeye:
Closes #430 

Decided to update some helpers so that a custom message can be passed. I then just passed a custom message to mention that the params have been updated, and that the event property was removed from the `_edItem` object param.

### DIFF
--- a/packages/eslint-plugin-pf-codemods/test/testHelpers.js
+++ b/packages/eslint-plugin-pf-codemods/test/testHelpers.js
@@ -277,14 +277,15 @@ function getInvalidSwapCallbackParamTests(
   componentNameArray,
   propNameArray,
   previousParamIndex,
-  newParamName
+  newParamName,
+  createMessageCallback
 ) {
   let tests = getInvalidAddCallbackParamTests(
     componentNameArray,
     propNameArray,
     newParamName,
     previousParamIndex,
-    getAddCallbackParamMessage
+    createMessageCallback
   );
 
   const formattedNewParamName =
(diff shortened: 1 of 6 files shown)


packages/eslint-plugin-pf-codemods/test/testHelpers.js
```diff
@@ -278,7 +283,8 @@ function getInvalidSwapCallbackParamTests(
     componentNameArray,
     propNameArray,
     newParamName,
-    previousParamIndex
+    previousParamIndex,
+    getAddCallbackParamMessage
```
wise-king-sullyman, line 287: I'm a bit confused as to why you didn't add your `createMessageCallback` here?

thatblindgeye: Mainly because I was only utilizing the add callback portion of these helpers rather than the swap callback. Would make sense to allow customizing a message for swap, though, so I can push an update to add that in

wise-king-sullyman (approved):

gitdallas (approved):

wise-king-sullyman:
@thatblindgeye just needs a rebase

wise-king-sullyman (approved):

PR #466, opened 2023-05-11
feat(Datalist): provide fixer for selectableRow prop

### LINKED ISSUE #368: DataList - create a fixer can be made for selectableRow change
Follow up to #366 

Ideally we would be able to provide a fixer for this change.

_Required actions:_
1. Update codemod to include fixer
2. Update tests
3. Update readme description & example

adamviktora:
Closes #368

One thing I don't like about this fixer is that when having an object inside the `selectableRow` prop like this:

```
const selectableRowObject = { onChange: (id, event) => {}};
<DataList selectableRow={selectableRowObject} />
```

It does not change the `selectableRowObject` variable name, but changes its type from an object to a function like this:
```
const selectableRowObject = (event, id) => {};
<DataList onSelectableRowChange={selectableRowObject} />
```

I mean, it could be renamed but at the same time we should not rename consumers variables. After writing the fixer, I get why this is one of the codemods which is better to be repaired manually.

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/datalist-remove-selectableRow.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/datalist-remove-selectableRow.js
@@ -27,19 +27,7 @@ module.exports = {
     const handleObjectExpression = (expr, toReplace) => {
       const propVal = getObjectProp(expr, "onChange")?.value;
 
-      if (!propVal) {
-        return;
-      }
-
-      if (propVal.type === "Identifier") {
-        toReplace.push([expr, sourceCode.getText(propVal)]);
-        handleIdentifierOfFunction(propVal, toReplace);
-      } else if (
-        ["ArrowFunctionExpression", "FunctionExpression"].includes(
-          propVal.type
-        ) &&
-        [1, 2].includes(propVal.params.length)
-      ) {
+      if (propVal) {
         toReplace.push([expr, sourceCode.getText(propVal)]);
       }
     };
@@ -72,7 +60,6 @@ module.exports = {
             variable?.defs[0].node.parent.range[0]
           );
           toReplace.push([n, sourceCode.getText(n.object)]);
-          handleRightAssignmentSide_Function(n.parent.right, toReplace);
         } else if (
           n.type === "AssignmentExpression" &&
           n.left.name === identifier.name
@@ -84,40 +71,6 @@ module.exports = {
       const variableValue = variable?.defs[0].node.init;
       variableValue &&
         handleRightAssignmentSide_Object(variableValue, toReplace);
-    };
-
-    const handleRightAssignmentSide_Function = (node, toReplace) => {
-      if (node.type === "Identifier") {
-        handleIdentifierOfFunction(node, toReplace);
-      }
-    };
-
-    const handleIdentifierOfFunction = (identifier, toReplace) => {
-      const variable = findVariableDeclaration(
-        identifier.name,
-        sourceCode.getScope(identifier)
-      );
-
-      const variableType = variable?.defs[0].type;
-      const variableNode = variable?.defs[0].node;
-
-      if (variableType === "Variable") {
-        const nodesUsingVariable = variable.references.map(
-          (ref) => ref.identifier.parent
-        );
-
-        nodesUsingVariable.forEach((n) => {
-          if (
-            n.type === "AssignmentExpression" &&
-            n.left.name === identifier.name
-          ) {
-            handleRightAssignmentSide_Function(n.right, toReplace);
-          }
-        });
-
-        variableNode.init &&
-          handleRightAssignmentSide_Function(variableNode.init, toReplace);
-      }
     };
 
     return {
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/datalist-remove-selectableRow.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/datalist-remove-selectableRow.js
@@ -149,27 +149,13 @@ module.exports = {
 
         toReplace.push([selectableRowAttr.name, "onSelectableRowChange"]);
 
-        // Having the same fix more than once would lead to errors
-        let toReplaceUnique = [];
-
-        for (const [oldVal, newVal] of toReplace) {
-          if (
-            toReplaceUnique.every(
-              ([oldUnique, newUnique]) =>
-                oldUnique !== oldVal || newUnique !== newVal
-            )
-          ) {
-            toReplaceUnique.push([oldVal, newVal]);
-          }
-        }
-
-        toReplaceUnique.length &&
+        toReplace.length &&
           context.report({
             message: `DataList's selectableRow property has been replaced with onSelectableRowChange. The order of the params in the callback has also been updated so that the event param is first.`,
             node,
             fix: function (fixer) {
               return [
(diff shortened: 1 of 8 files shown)


packages/eslint-plugin-pf-codemods/lib/rules/v5/datalist-remove-selectableRow.js
```diff
@@ -1,25 +1,244 @@
...
+
+      if (expr.type === "Identifier") {
+        handleIdentifierOfObject(expr, toReplace);
+      }
+
+      toReplace.push([selectableRowAttr.name, "onSelectableRowChange"]);
+    }
+
+    // Having the same fix more than once would lead to errors
+    let toReplaceUnique = [];
```
thatblindgeye, line 196: (think I accidentally deleted the wrong thing)

Are there any particular fixes that are being duplicated within toReplace? Just wondering if needing to have another array just to get the unique toReplace items means the fix for the original issue is too complicated or anything.

packages/eslint-plugin-pf-codemods/lib/rules/v5/datalist-remove-selectableRow.js
```diff
@@ -1,25 +1,244 @@
...
+      if (!propVal) {
+        return;
+      }
+
+      if (propVal.type === "Identifier") {
+        toReplace.push([expr, sc.getText(propVal)]);
+        handleIdentifierOfFunction(propVal, toReplace);
+      } else if (propVal.type === "ArrowFunctionExpression") {
+        const paramsLen = propVal.params.length;
+        if (1 <= paramsLen <= 2) {
```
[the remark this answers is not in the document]
adamviktora: Oh, didn't know that, I am used to doing this in Python, where it works logically fine

packages/eslint-plugin-pf-codemods/lib/rules/v5/datalist-remove-selectableRow.js
```diff
@@ -1,25 +1,244 @@
...
+    const sc = context.getSourceCode();
+
+    const getObjectProp = (expr, propName) =>
+      expr.properties.find(
+        (prop) =>
+          (prop.key.type === "Identifier" && prop.key.name === propName) ||
+          (prop.key.type === "Literal" && prop.key.value === propName)
+      );
+
+    const swapCallbackParam = (params, toReplace) => {
```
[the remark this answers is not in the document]
adamviktora: Good idea, that didn't come to my mind, that we can run some rules before others. This rule can be a good case for separating the callback swap into another rule. Because currently I use not only `swapCallbackParam`, but also some not very nice string handling in `handleObjectExpression`, where instead of just replacing object like this ` { onChange: function (id, event) {} };` with the `function (id, event) {}`, I am swapping the params right away, which needs to be done with the not so nice combination of strings.

Working on a refactoring right now to make the swapping as a separate rule.

packages/eslint-plugin-pf-codemods/lib/rules/v5/datalist-remove-selectableRow.js
adamviktora: I am testing it on this file:

```
import * as React from "react";
import { DataList } from "@patternfly/react-core";

const onChange = (id) => {};
function onChangeFunc(id, event) {}

const selectableRowObject = { onChange: function (id, event) {} };
selectableRowObject.onChange = onChangeFunc;
selectableRowObject["onChange"] = onChange;

let selectableRowObject2 = { onChange: onChangeFunc };
selectableRowObject2 = selectableRowObject;

<>
  <DataList selectableRow={{ onChange: (id, event) => {} }} />
  <DataList selectableRow={{ onChange: onChangeFunc }} />
  <DataList selectableRow={{ onChange }} />

  <DataList selectableRow={selectableRowObject} />
  <DataList selectableRow={selectableRowObject2} />
</>;
```

Problem is when `onChangeFunc` gets referenced more than once. If you comment out this part:
```
<DataList selectableRow={selectableRowObject} />
<DataList selectableRow={selectableRowObject2} />
```
It will work without the need of `toReplaceUnique`, otherwise it breaks (don't really remember why right now, but I guess when attempting to replace a token which has already been replaced).

packages/eslint-plugin-pf-codemods/lib/rules/v5/datalist-remove-selectableRow.js
```diff
@@ -1,25 +1,173 @@
...
+            message: `DataList's selectableRow property has been replaced with onSelectableRowChange. The order of the params in the callback has also been updated so that the event param is first.`,
+            node,
+            fix: function (fixer) {
+              return [
+                ...toReplace.map(([oldValue, newValue]) =>
+                  fixer.replaceText(oldValue, newValue)
+                ),
+                ...[...new Set(rangeStartsToReplaceConst)].map((rangeStart) =>
+                  fixer.replaceTextRange(
+                    [rangeStart, rangeStart + "const".length],
```
thatblindgeye, line 163: Any reason for this instead of just `rangeStart + 5`?

packages/eslint-plugin-pf-codemods/lib/rules/v5/datalist-remove-selectableRow.js
```diff
@@ -1,25 +1,173 @@
...
-            context.report({
-              node,
-              message: `DataList's selectableRow property has been replaced with onSelectableRowChange. The order of the params in the callback has also been updated so that the event param is first.`,
-            });
-          }
+    if (!dataListImport) {
+      return {};
+    }
+
+    let rangeStartsToReplaceConst = [];
```
thatblindgeye, line 16: Are there instances where this will ever have a length larger than 1, or have duplicate items where the `new Set` on line 161 is needed?

thatblindgeye (commented):
This is looking nicer 💪🏼 Had a couple more nits/comments below

packages/eslint-plugin-pf-codemods/lib/rules/v5/datalist-remove-selectableRow.js
adamviktora: Yes, this can have length larger than one. It is filled every time when we are changing the value of `.onChange` prop of the `selectableRowObject` (so in a `MemberExpression`). And because we are changing the original variable `const` object to a function, we have to change it to `let`, as then we assing a new value not to the `.onChange` prop, but to the function variable directly.

Example:
```
const onChange = (id) => {};
function onChangeFunc(id, event) {}

const selectableRowObject = { onChange: function (id, event) {} };
selectableRowObject.onChange = onChangeFunc;
selectableRowObject["onChange"] = onChange;
selectableRowObject["onChange"] = onChangeFunc;

const selectableRowObject2 = { onChange: function (id, event) {} };
selectableRowObject2.onChange = onChangeFunc;

<>
  <DataList selectableRow={selectableRowObject} />
  <DataList selectableRow={selectableRowObject2} />
</>;
```
console.log():
```
rangeStartsToReplaceConst
[ 150, 150, 150, 355 ]
```
without deduplication with the `new Set` it throws an error, as we are trying to replace it more times.

There will probably not be many (maybe none) cases, where someone uses the selectableRowObject like that and reassigns the value, but just to be sure

gitdallas (approved):
lgtm, thanks for the helper as well

wise-king-sullyman (approved):
🥳

PR #488, opened 2023-05-23
fix(EmptyState): account for missing variant case

### LINKED ISSUE #485: EmptyState - string variant instances are not getting updated
Following codemods 1.76.0 testing - this is throwing errors in Keycloak.

There are several instances where the variant prop is set to a string ie `variant="large"` but the codemod only accounts for instances that match `variant={EmptyStateVariant.size}`

jenny-s51:
Closes #485 - adds codemod fix for when `variant` is set to a string


packages/eslint-plugin-pf-codemods/lib/rules/v5/emptyStateVariant-rename-values.js
```diff
@@ -2,14 +2,26 @@ const { getFromPackage } = require("../../helpers");
...
+      "@patternfly/react-core"
+    ).imports.find((specifier) => specifier.imported.name === "EmptyState");
+
+
+    const rename = {
+      small: "sm",
+      large: "lg",
+    };
+
+    return !emptyStateVariantImport|| !emptyStateImport
```
thatblindgeye, line 24: This is causing the new test to fail since `emptyStateVariantImport` is false in it.

```suggestion
    return !emptyStateVariantImport && !emptyStateImport
```

packages/eslint-plugin-pf-codemods/lib/rules/v5/emptyStateVariant-rename-values.js
```diff
@@ -39,6 +51,38 @@ module.exports = {
...
+          JSXOpeningElement(node) {
+            if (emptyStateImport.local.name === node.name.name) {
+              const variantProp = node.attributes.find(
+                (attribute) => attribute.name?.name === "variant"
+              );
+
+              if (
+                variantProp &&
+                (rename[variantProp.value?.value] ||
+                  variantProp.value?.type === "Literal")
```
thatblindgeye, line 63: We should be able to omit the type === Literal portion as that seems to be causing a valid test to expect an error since `variant="lg"` causes the type === Literal to be true.

```suggestion
                rename[variantProp.value?.value]
```

packages/eslint-plugin-pf-codemods/test/rules/v5/emptyStateVariant-rename-values.js
```diff
@@ -56,5 +60,17 @@ ruleTester.run("emptyStateVariant-rename-values", rule, {
...
     },
+    {
+      code: `import { EmptyState } from '@patternfly/react-core/dist/esm/components/EmptyState/index.js'; 
+      <EmptyState variant="large" />`,
+      output: `import { EmptyState } from '@patternfly/react-core/dist/esm/components/EmptyState/index.js'; 
+      <EmptyState variant="lg" />`,
+      errors: [
+        {
+          message: `The "variant" prop type for EmptyState has been updated. The previous values of "small" and "large" have been replaced with values of "sm" and "lg" respectively.`,
+          type: "MemberExpression",
```
thatblindgeye, line 71:
```suggestion
          type: "JSXOpeningElement",
```

jenny-s51 (commented):
Thank you for your helpful suggestions @thatblindgeye - tests are passing 🎉

thatblindgeye (approved):

gitdallas (approved):

PR #496, opened 2023-05-24
fix(Popover): fix issue with identifier handlers, remove fixers

### LINKED ISSUE #495: Popover - popover-remove-props
Encountered the following error when running codemods against @patternfly/quickstarts:

image (10)

NOTE: Manually removing the `onShow` and `onHide` props from the component usage and re-running pf-codemods remedies the problem.

gitdallas:
closes #495 

removes fixers as a developer will have to remove instances of parameter usage as well as the parameter variable itself

fixes issue where using a function variable wasn't working, like

```
const myFunc = (fn) => {};
<Popover onHide={myFunc} />
```

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/popover-remove-props.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/popover-remove-props.js
@@ -50,20 +50,30 @@ module.exports = {
             const attr = node.openingElement.attributes.find(attribute => attribute.name?.name === name)
             if(attr) {
               const funct = attr.value.expression;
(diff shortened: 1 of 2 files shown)


wise-king-sullyman (approved):

thatblindgeye (approved):
Looks good. Not something caused by this PR, but I am noticing the `popover-remove-props` rule and the `popover-swap...` rule are conflicting for the shouldClose prop. In the test file at least, the `shouldClose={(foo, event)...` gets updated so that foo is removed, leaving the prop as `shouldClose={(event)....` rather than swapping them.

Could open a separate issue unless you'd want to try doing a quick fix here.

PR #501, opened 2023-05-30
Removed aliasSuffixing from path update codemods

### LINKED ISSUE #365: Improve moveSpecifiers helper and test helper (not a codemod)
To try improving these helpers further:

- Take into account imports for things such as interfaces. Right now an item like `{name: "ComponentProps, type: "props"}` most likely wouldn't be updated in consumer code for a typical use case (the helper is only searching for non-component imports being passed to some component prop), and would be tested by the test helper passing it in like, `prop={ComponentProps}`, which wouldn't really be accurate. 
- For the invalid test helper, we should probably add a param for aliasSuffix so that we could test components that aren't deprecated. Right now it's hardcoded to alias any non-component imports which isn't always necessary.

### LINKED ISSUE #500: Bug - v5 alpha testing - Missing Select* imports
Testing Cost Management and the latest codemods. Imports for SelectOptionObject and SelectDirection are being removed without replacement.

Screenshot 2023-05-26 at 9 25 37 AM

thatblindgeye:
Closes #500, closes #365 (no longer applicable with this update)

Also update the logic to prevent TableComposable being renamed to Table and then being moved to deprecated so that it matches the Export logic (just add a comment next to the import specifier).


packages/eslint-plugin-pf-codemods/test/rules/v5/table-rename-TableComposable.js
```diff
@@ -17,7 +17,7 @@ ruleTester.run("table-rename-TableComposable", rule, {
   invalid: [
     {
       code: `import { TableComposable } from '@patternfly/react-table'; <TableComposable>Body</TableComposable>`,
-      output: `import { Table } from '@patternfly/react-table'; <Table data-codemods="true">Body</Table>`,
+      output: `import { Table /* data-codemods */ } from '@patternfly/react-table'; <Table>Body</Table>`,
```
wise-king-sullyman, line 20: Is the `/* data-codemods */` in the output here the intended behavior?

thatblindgeye: Yeah, I updated how we track what imports to not update paths for with the comment (instead of the data attribute).

wise-king-sullyman (approved):

packages/eslint-plugin-pf-codemods/test/rules/v5/table-rename-TableComposable.js
gitdallas: would we want to note this in the readme doc?

packages/eslint-plugin-pf-codemods/lib/helpers.js
```diff
@@ -4,37 +4,26 @@ function moveSpecifiers(
...
-      const hasDataAttr = foundElement?.openingElement.attributes?.find(
-        (attr) => attr.name?.name === "data-codemods"
-      );
+      getFromPackage(context, fromPackage, specifiersToMove);
 
-      return (foundElement && !hasDataAttr) || !foundElement;
-    });
+    const getSpecifiersToMove = (specifiers) =>
+      specifiers.filter((specifier) => {
+        const comments = src.getCommentsAfter(specifier);
```
gitdallas, line 16: can we add some tests that test `/* data-codemods */` both for valid (places where we don't want to operate on them if they exist) and invalid (like we have a `/* data-codemods */` on a different specifier and it doesn't prevent the codemod from working on the component specifier we are working with)?

gitdallas (changes requested):

gitdallas (approved):
🚀

PR #503, opened 2023-05-31
feat(Next components): updated import and export paths

### LINKED ISSUE #471: Next components - update import paths
Possible followup to #440. Currently we have codemods for updating the path for deprecated components to our deprecated directory, but do not have any for updating the path for "next" components to just `@patternfly/react-core`.  Might be more of a "nice to have" codemod depending on how common the usage is for our `@patternfly/react-core/next` components.

_Required actions:_
1. Build codemod
2. Build test

thatblindgeye:
Closes #471

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/helpers.js
+++ b/packages/eslint-plugin-pf-codemods/lib/helpers.js
@@ -35,12 +35,22 @@ function moveSpecifiers(
       if (firstSpecifier?.parent?.source?.value?.includes("dist/esm")) {
         //expecting @patternfly/{package}/{designator} where designator is next/deprecated
         const toParts = toPackage.split("/");
-        //expecting @patternfly/{package}/dist/esm/components/{Component}/index.js
-        //needing toPath to look like fromPath with the designator before /components
         const fromParts = firstSpecifier.parent.source.value.split("/");
-        if (toParts[0] === "@patternfly" && toParts.length === 3) {
-          fromParts.splice(4, 0, toParts[2]);
-          modifiedToPackage = fromParts.join("/");
+        if (toParts[0] === "@patternfly") {
+          //expecting @patternfly/{package}/dist/esm/components/{Component}/index.js
+          //needing toPath to look like fromPath with the designator before /components
+          if (toParts.length === 3) {
+            fromParts.splice(4, 0, toParts[2]);
+            modifiedToPackage = fromParts.join("/");
+          }
+
+          // Expecting @patternfly/{package}/dist/esm/next/components/{Component}/index.js
+          // Needing toPath to look like fromPath *without* the designator before /components
+          if (toParts.length === 2) {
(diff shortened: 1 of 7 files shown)


packages/eslint-plugin-pf-codemods/lib/helpers.js
```diff
@@ -35,12 +35,22 @@ function moveSpecifiers(
...
+            modifiedToPackage = fromParts.join("/");
+          }
+
+          // Expecting @patternfly/{package}/dist/esm/next/components/{Component}/index.js
+          // Needing toPath to look like fromPath *without* the designator before /components
+          if (toParts.length === 2) {
+            modifiedToPackage = fromParts
+              .filter((part) => part !== "next")
+              .join("/");
+          }
```
wise-king-sullyman, line 53: I wouldn't block over it, but I wonder if this logic could be reworked a bit so that we can avoid getting this deep in nested if statements. IMO nested if statements are (at least most of the time) the kind of thing that increases cognitive load when trying to reason about code exponentially with each level.

Simplest way to do it in this case would probably be to just refactor to have guard clauses for the preceding checks. But there very well could be a better way than that.

packages/eslint-plugin-pf-codemods/test/testHelpers.js
```diff
@@ -434,34 +428,67 @@ function getMoveSpecifiersValidtests(specifiersToMoveArray) {
...
-      code: `export { ${specifier} as CustomExport } from '@patternfly/react-${package}/dist/esm/deprecated/components/${specifier}/index.js';`,
-    });
     // No @patternfly/react-core export
     tests.push({
       code: `export { ${specifier} as CustomExport } from 'foo';`,
     });
+
+    if (movingFrom === "deprecated") {
+      tests.push({
+        code: `import { ${specifier} } from '@patternfly/react-${package}/next';`,
```
wise-king-sullyman, line 438: I very likely am just missing something obvious here, but why are you testing the next subDir if it's being moved from deprecated?

thatblindgeye: That was added a bit ago, but I think it was just to check that something from the /next directory wasn't going to be deprecated. Probably not necessary so I don't mind removing it if we don't really need it.

packages/eslint-plugin-pf-codemods/lib/helpers.js
thatblindgeye: Yeah I agree, I'll rework this this afternoon since it should be simple enough of a fix.

packages/eslint-plugin-pf-codemods/test/testHelpers.js
wise-king-sullyman: Ah ok, that makes sense. I'm not opposed to keeping it.

PR #518, opened 2023-06-08
fix(move specifiers) keep inputkind while moving specifiers

### LINKED ISSUE #512: 'type' keyword not preserved when updating path
When importing with `type`, the keyword gets removed after updating the path via moveSpecifiers.

eg.
`import type { someType } from "@patternfly/react-core"`  turns into
`import { someType } from "@patternfly/react-core/deprecated"`

gitdallas:
closes #512

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/helpers.js
+++ b/packages/eslint-plugin-pf-codemods/lib/helpers.js
@@ -12,7 +12,8 @@ function moveSpecifiers(
       getFromPackage(context, fromPackage, specifiersToMove);
 
     const getSpecifiersToMove = (specifiers) =>
-      specifiers.filter((specifier) => {
+    {
+      const specs = specifiers.filter((specifier) => {
         const comments = src.getCommentsAfter(specifier);
 
         return (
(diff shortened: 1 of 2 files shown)


thatblindgeye (approved):

packages/eslint-plugin-pf-codemods/lib/helpers.js
```diff
@@ -74,7 +74,10 @@ function moveSpecifiers(
         return (
           node?.type === nodeType &&
           [modifiedPackage, toPackage].includes(node?.source?.value) &&
-          specifierReference?.parent?.importKind === node?.importKind
+          (
+            specifierReference?.parent?.importKind === node?.importKind &&
+            specifierReference?.parent?.exportKind === node?.exportKind
```
gitdallas, line 79: i figured i'd just do && here rather than checking if we are dealing with import or export and then checking that one... this works because if we are dealing with an import... exportKind will be undefined in both places and vice versa

thatblindgeye (approved):

PR #525, opened 2023-06-26
fix(CardHeader): updated logic to take into account conditional rendering

### LINKED ISSUE #524: CardHeader - API update not working with conditional rendering
The `cardHeader-update-api` codemod doesn't work as intended when CardActions is conditionally rendered (most likely the same for CardHeaderMain). A structure like the following won't be updated (though the imports *will* be removed):

```
<CardHeader>
   {dropdownItems && <CardActions />}
</CardHeader>
```

thatblindgeye:
Closes #524 

@gitdallas this runs into the possible issue that the test file has a "first run" and "second run" test due to the way the fix works. Testing in the `test.tsx` file things look to work as intended, though.

### DIFF
--- a/packages/eslint-plugin-pf-codemods/lib/rules/v5/cardHeader-update-api.js
+++ b/packages/eslint-plugin-pf-codemods/lib/rules/v5/cardHeader-update-api.js
@@ -8,142 +8,146 @@ module.exports = {
     const removedCardImports = imports.filter((specifier) =>
       ["CardHeaderMain", "CardActions"].includes(specifier.imported.name)
     );
-    const cardHeaderImport = imports.find((specifier) => specifier.imported.name === "CardHeader");
+    const cardHeaderImport = imports.find(
+      (specifier) => specifier.imported.name === "CardHeader"
+    );
+
+    const findParents = (node, elementNameToFind) => {
+      const parents = {
+        current: node?.parent,
+        conditionalExpression: undefined,
+      };
+
+      while (parents.current) {
+        if (parents.current.openingElement?.name?.name === elementNameToFind) {
+          return parents;
+        }
+        if (
+          ["ConditionalExpression", "LogicalExpression"].includes(
+            parents.current.type
+          )
+        ) {
+          parents.conditionalExpression = parents.current;
+        }
+
+        parents.current = parents.current.parent;
+      }
+
+      return undefined;
+    };
+    const getComponentContent = (component) => {
+      const tagsRegex = new RegExp(
+        `<\/?${component.openingElement.name.name}(.*?)>`,
+        "g"
+      );
+
+      return context.getSourceCode().getText(component).replace(tagsRegex, "");
+    };
+    const createReplacementContent = (content, conditionOrLogicExpression) => {
+      const isCardHeaderMainContent = /^<>.*<\/>$/.test(content);
+
+      if (!conditionOrLogicExpression) {
+        return isCardHeaderMainContent ? content : `actions={${content}}`;
+      }
+
+      const sourceCode = context.getSourceCode();
+
+      if (conditionOrLogicExpression.type === "LogicalExpression") {
+        const logicalExpressionContent = sourceCode.getText(
+          conditionOrLogicExpression.left
+        );
+        const { operator } = conditionOrLogicExpression;
+
+        return isCardHeaderMainContent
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin-pf-codemods/lib/rules/v5/cardHeader-update-api.js
```diff
@@ -8,142 +8,146 @@ module.exports = {
...
+        return isCardHeaderMainContent
+          ? `{${conditionalTestContent} ? ${content} : ${conditionalAlternateContent}}`
+          : `actions={${conditionalTestContent} ? ${content} : {}}`;
+      }
+    };
 
     return ![...removedCardImports, cardHeaderImport].length
       ? {}
       : {
-          ImportDeclaration(node) {
```
thatblindgeye, line 16: Removing the imports should be taken care of by the no unused import rule, so we don't really need to handle that logic here.

packages/eslint-plugin-pf-codemods/lib/rules/v5/cardHeader-update-api.js
```diff
@@ -8,142 +8,146 @@ module.exports = {
     const removedCardImports = imports.filter((specifier) =>
       ["CardHeaderMain", "CardActions"].includes(specifier.imported.name)
     );
-    const cardHeaderImport = imports.find((specifier) => specifier.imported.name === "CardHeader");
+    const cardHeaderImport = imports.find(
+      (specifier) => specifier.imported.name === "CardHeader"
+    );
+
+    const findParents = (node, elementNameToFind) => {
```
gitdallas, line 15: wonder if this should be a helper func

packages/eslint-plugin-pf-codemods/lib/rules/v5/cardHeader-update-api.js
```diff
@@ -8,142 +8,146 @@ module.exports = {
...
+      const parents = {
+        current: node?.parent,
```
gitdallas, line 17: current makes sense for the loop internal to this function but i'm not sure it makes sense as the return type - maybe more important if we did make this a helper func

also, should it be findParent() or findClosestParent() maybe as we are only going to return as soon as we find what we are looking for and not have an array of parents... or i suppose plural is because you are also finding the expression as a parent.. 

i suppose i'm showing my thoughts as i'm thinking and reading through this. it makes sense to me now, might make sense to just be overly verbose `getExpressionAndParentCardHeader` as that is all we are doing with this function and i think i'd go with `parent` instead of `current`.

but really all these words are just for a nit at this point 🤣

packages/eslint-plugin-pf-codemods/lib/rules/v5/cardHeader-update-api.js
```diff
@@ -8,142 +8,146 @@ module.exports = {
...
+        ) {
+          parents.conditionalExpression = parents.current;
+        }
+
+        parents.current = parents.current.parent;
+      }
+
+      return undefined;
+    };
+    const getComponentContent = (component) => {
```
gitdallas, line 38: is there no `innerHTML` or equivalent that could have achieved the same?

gitdallas (approved):
looks thorough and good, if anything maybe it could be a little more readable.. but pretty impressive you were able to whip this up as quick as you did

packages/eslint-plugin-pf-codemods/lib/rules/v5/cardHeader-update-api.js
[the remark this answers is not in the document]
thatblindgeye: Yeah I considered it, but for now since it's for this specific use case didn't think we really needed it.

packages/eslint-plugin-pf-codemods/lib/rules/v5/cardHeader-update-api.js
thatblindgeye: Makes sense, I updated the naming of things for this. Let me know what you think

packages/eslint-plugin-pf-codemods/lib/rules/v5/cardHeader-update-api.js
thatblindgeye: Replaced it with just mapping over the node children to `getText` instead

wise-king-sullyman (approved):

PR #562, opened 2024-02-07
chore(build): add TS support

### LINKED ISSUE #561: Add TS support
Part of #554

wise-king-sullyman:
Closes #561 

This does:
- Add basic build infrastructure to support TS usage in the repo
- Update the entire class-name-updater package to TS
- Update the operating logic of eslint-plugin-pf-codemods to TS

This does not:
- Update the pf-codemods package to TS, because I plan to make a new package for v6 forward to prevent breaking the pf-codemods package for v5 updates
- Update the rules or tests in eslint-plugin-pf-codemods to TS, because we don't plan to do development on the existing rules. New rules created for, or rules copied to, v6 forward should be TS from the start.


gitdallas (approved):
👍

thatblindgeye (approved):

PR #565, opened 2024-02-14
chore(refactor): Refactor rule structure to prevent merge conflicts

### LINKED ISSUE #554: Prepare codemods repo for v6
Create scaffolding and make tooling changes needed for us to begin v6 codemod work.

wise-king-sullyman:
Closes #554


packages/pf-codemods/v5-README.md
```diff
@@ -0,0 +1,98 @@
...
+  -h, --help         display help for command
+```
+
+## Rules
+
+These rules are based off the breaking change notes for React. Each rule links the breaking change patternfly-react PR in case you want to better understand the change. Also, each rule makes sure you're using a PatternFly component before running.
+
+Some rules will add either a comment (`/* data-codemods */`) or data attribute (`data-codemods="true"`) in order to prevent certain other rules from applying an unnecessary fix.
+
+### example-rule[(#1234)](https://github.com/patternfly/patternfly-react/pull/1234)
```
thatblindgeye, line 49: This README needs to be fixed to house the v5 content

packages/pf-codemods/index.js
```diff
@@ -55,67 +63,84 @@ async function printResults(eslint, results, format) {
...
 
+function ruleVersion(options) {
+  if (options.v4) {
+    return [...ruleVersionMapping["v5"], ...ruleVersionMapping["v6"]];
+  } else if (options.v6) {
+    return [...ruleVersionMapping["v4"], ...ruleVersionMapping["v5"]];
+  } else {
+    return [...ruleVersionMapping["v4"], ...ruleVersionMapping["v6"]];
+  }
+}
```
thatblindgeye, line 81: Just to confirm, the point of this is to return the rules thare *aren't* for the version specified in options? So `v4` in options means we want this to return the v5 and v6 rules so that we can remove them?

If so then a slight nit would be the function name; Reading this at first I thought this would be an array of the rules we'd want to run, but further down looks like these are the ones we're excluding.

packages/pf-codemods/scripts/buildReadme.ts
```diff
@@ -0,0 +1,42 @@
...
+const builtReadMes = v6RuleDirs.map(async (ruleDir) => {
+  const files = await readdir(ruleDir);
+
+  const readMe = files.find((file) => file.includes(".md"));
+  const input = files.find((file) => file.includes("Input.tsx"));
+  const output = files.find((file) => file.includes("Output.tsx"));
+
+  if (!readMe || !input || !output) {
+    console.error("Missing file in rule directory: ", ruleDir);
+    throw new Error("Missing file in rule directory");
```
thatblindgeye, line 23: Do we need both the console and throw error, or could we just include the ruleDir in the thrown error message?

thatblindgeye (commented):
Really only one thing that needs to be updated (v5 README), otherwise questions/possible updates.

Also as potential followups that we went over during our meet up:

- A mass test file for input/output where all rule example are in a single file (essentially the `test.tsx` file in the test directory.
- A table of contents for the rules in the README to more quickly jump to a particular rule (rather than having to scroll/find in the page)

packages/pf-codemods/v5-README.md
wise-king-sullyman: Whoops

packages/pf-codemods/index.js
wise-king-sullyman: Yeah that was how I thought it worked initially before as well, I'll see what I can come up with for better name.

packages/pf-codemods/scripts/buildReadme.ts
wise-king-sullyman: Hmmmmm probably

thatblindgeye (approved):

PR #568, opened 2024-02-16
chore(Page): updated props  @thatblindgeye

### LINKED ISSUE #543: Page - Penta updates
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/9848 and https://github.com/patternfly/patternfly-react/pull/9774

Added a new option to the `variant` prop, and updated so the class from the `variant` prop will only be applied when `type="default"` (previously a class would always be applied)

PageSidebar: Removed Page sidebar theme prop since it only supports the default theme.
PageSection: Removed light, dark and `darker variant since the are no longer supported.
PageHeader (deprecated): removed isSelected prop since it is no longer supported in Page

_Required actions:_
1. Build codemod
2. Build test

thatblindgeye:
Closes #543

### DIFF
--- a/generators/src/write-rule.ts
+++ b/generators/src/write-rule.ts
@@ -31,7 +31,7 @@ export async function genericRule({
 // https://github.com/patternfly/patternfly-react/pull/${referencePR}
 module.exports = {
   meta: { fixable: 'code' },
-  create: function(context: { report: (arg0: { node: any; message: string; fix(fixer: any): any; }) => void; }) {
+  create: function(context: { report: (arg0: { node: any; message: string; fix?(fixer: any): any; }) => void; }) {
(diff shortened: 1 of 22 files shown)


generators/src/write-rule.ts
```diff
@@ -31,7 +31,7 @@ export async function genericRule({
 // https://github.com/patternfly/patternfly-react/pull/${referencePR}
 module.exports = {
   meta: { fixable: 'code' },
-  create: function(context: { report: (arg0: { node: any; message: string; fix(fixer: any): any; }) => void; }) {
+  create: function(context: { report: (arg0: { node: any; message: string; fix?(fixer: any): any; }) => void; }) {
```
thatblindgeye, line 34: Made the fix param optional since we don't need it for warning rules.

wise-king-sullyman (approved):

PR #570, opened 2024-02-19
feat(SimpleFileUpload): added rule to warn about component changes

### LINKED ISSUE #551: Simple file upload - Penta updates
Follow up to breaking change PR:

- https://github.com/patternfly/patternfly-react/pull/10026
  - removed the `aria-describedby` from the TextInput component
  - removed the `id` attribute from the browse button (was only being used for the above `aria-describedby`
  - example(s) using FormHelperText now just use FileUploadHelperText as a container (provide a warning that consumers should make this update)

_Required actions:_
1. Build codemod
2. Build test

thatblindgeye:
Closes #551

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/simpleFileUploadWarnChanges/simpleFileUploadWarnChangesInput.tsx
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/simpleFileUploadWarnChanges/simpleFileUploadWarnChangesInput.tsx
@@ -1,0 +1,2 @@
+import { SimpleFileUpload } from "@patternfly/react-core";
+import { SomethingElse } from "@patternfly/react-core";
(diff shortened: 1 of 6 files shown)


packages/eslint-plugin-pf-codemods/src/rules/v6/simpleFileUploadWarnChanges/simpleFileUploadWarnChangesInput.tsx
```diff
@@ -0,0 +1,2 @@
+import { SimpleFileUpload } from "@patternfly/react-core";
```
wise-king-sullyman, line 1: Did you mean to remove the input and output files?

wise-king-sullyman (approved):
One comment, not blocking though.

PR #571, opened 2024-02-19
feat(Avatar) replace border with isBordered

### LINKED ISSUE #549: Avatar - Penta updates
Follow up to breaking change PR  https://github.com/patternfly/patternfly-react/pull/9954

`border` prop has been removed since light and dark borders are no longer supported.  `isBordered` prop was added to add a border to avatar

_Required actions:_
1. Build codemod
2. Build test

Dominik-Petrik:
closes #549

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/avatarReplaceBorderForIsBordered/avatar-replace-border-for-isBordered.md
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/avatarReplaceBorderForIsBordered/avatar-replace-border-for-isBordered.md
@@ -1,6 +1,9 @@
-### avatar-replace-border-for-isBordered [(#9954)](https://github.com/patternfly/patternfly-react/pull/9954)
+### avatar-replace-border-for-isBordered
 
-border prop has been removed since light and dark borders are no longer supported. isBordered prop was added to add a border to avatar.
+[(#9881)](https://github.com/patternfly/patternfly-react/pull/9881)
+[(#9954)](https://github.com/patternfly/patternfly-react/pull/9954)
+
+`border` prop has been removed from Avatar since theming is not longer handled React-side. We recommend using the `isBordered` prop instead to add a border to Avatar.
 
 #### Examples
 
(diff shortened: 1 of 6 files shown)


wise-king-sullyman (commented):
Code looks good (missing blank lines at the end of the files so I'm guessing you might want to run prettier on all of them, that's not blocking though).

One big question though: do you know what the deal is with all the yarn.lock changes? Seems odd when there is no package file change here.

Dominik-Petrik:
Hmm, let me investigate that

packages/eslint-plugin-pf-codemods/src/rules/v6/avatarReplaceBorderForIsBordered/avatar-replace-border-for-isBordered.md
```diff
@@ -0,0 +1,17 @@
+### avatar-replace-border-for-isBordered [(#9954)](https://github.com/patternfly/patternfly-react/pull/9954)
+
+border prop has been removed since light and dark borders are no longer supported. isBordered prop was added to add a border to avatar.
```
thatblindgeye, line 3: More of a nit, but we could rephrase the message to convey that border theming is no longer handled React-side (since dark/light borders are technically supported, just not something the consumer needs to really specify at the component level). Maybe something like:

```suggestion
`border` prop has been removed from Avatar since theming is not longer handled React-side. We recommend using the `isBordered` prop instead to add a border to Avatar.
```

packages/eslint-plugin-pf-codemods/src/rules/v6/avatarReplaceBorderForIsBordered/avatar-replace-border-for-isBordered.ts
```diff
@@ -0,0 +1,16 @@
+import { getFromPackage, renameProps } from '../../helpers';
+
+// https://github.com/patternfly/patternfly-react/pull/9954
```
thatblindgeye, line 3: Similar to the README comment, can we just add a link to the core bump PR in a comment as well

thatblindgeye (commented):
Few comments below, lmk what you think!

thatblindgeye (approved):
Other than yarn lock issue

PR #575, opened 2024-02-19
feat(JumpLinks): warned about href and markup/type updates

### LINKED ISSUE #559: Jump Links - Penta updates
- https://github.com/patternfly/patternfly-react/pull/10027#discussion_r1473517252
  - The `href` prop is now required

_Required actions:_
1. Build codemod
2. Build test

thatblindgeye:
Closes #559


packages/eslint-plugin-pf-codemods/src/ruleCustomization.ts
```diff
@@ -16,6 +16,7 @@ export const warningRules = [
   "emptyState-warn-change-structure",
   "formControls-updated-markup",
   "horizontalSubnav-warn-ariaLabel",
+  "jumpLinksItem-warn-prop-markup-updates",
```
wise-king-sullyman, line 19: I'm not sure if this should really be a warning since the `href` now being required will throw build errors.

packages/eslint-plugin-pf-codemods/src/rules/v6/jumpLinksItemWarnPropMarkupUpdates/jumpLinksItemWarnPropMarkupUpdatesInput.tsx
```diff
@@ -0,0 +1 @@
+import { JumpLinksItem } from "@patternfly/react-core";
```
wise-king-sullyman, line 1: Could these tsx files be deleted?

wise-king-sullyman (approved):
A couple non-blocking comments

packages/eslint-plugin-pf-codemods/src/rules/v6/jumpLinksItemWarnPropMarkupUpdates/jumpLinksItemWarnPropMarkupUpdatesInput.tsx
thatblindgeye: I think they'd be worth keeping in, just so that running the single tests locally we can still check that the output looks as expected (mainly warning vs error). They might be able to be used to create a mass test file where all the rules' input code is rendered to ensure no rules conflict with one another?

PR #580, opened 2024-02-20
feat(Label): replace isOverflowLabel with variant="overflow"

### LINKED ISSUE #556: Label - isOverflowLabel removed
`isOverflowLabel` has been removed, `variant="overflow"` is what it should be replaced with.

adamviktora:
Closes #556

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/labelRemoveIsOverflowLabel/label-remove-isOverflowLabel.md
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/labelRemoveIsOverflowLabel/label-remove-isOverflowLabel.md
@@ -1,6 +1,6 @@
 ### label-remove-isOverflowLabel [(#10037)](https://github.com/patternfly/patternfly-react/pull/10037)
 
-isOverflowLabel prop for Label has been replaced with variant="overflow"
+The `isOverflowLabel` prop for Label has been replaced with `variant="overflow"`. Running the fix for this rule will replace an existing `variant` prop (which had no effect if `isOverflowLabel` was used).
 
 #### Examples
 
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/labelRemoveIsOverflowLabel/label-remove-isOverflowLabel.md
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/labelRemoveIsOverflowLabel/label-remove-isOverflowLabel.md
@@ -1,0 +1,17 @@
+### label-remove-isOverflowLabel [(#10037)](https://github.com/patternfly/patternfly-react/pull/10037)
+
+isOverflowLabel prop for Label has been replaced with variant="overflow"
+
+#### Examples
+
+In:
+
+```jsx
+%inputExample%
+```
+
+Out:
+
+```jsx
+%outputExample%
+```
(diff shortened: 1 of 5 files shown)


packages/eslint-plugin-pf-codemods/src/rules/v6/labelRemoveIsOverflowLabel/label-remove-isOverflowLabel.test.ts
```diff
@@ -0,0 +1,24 @@
...
+ruleTester.run("label-remove-isOverflowLabel", rule, {
+  valid: [
+    {
+      code: `<Label isOverflowLabel />`
+    },
+    {
+      code: `import { Label } from '@patternfly/react-core'; <Label someOtherProp />`
+    }
+  ],
+  invalid: [
```
thatblindgeye, line 13: This will require an update to be made either to the renameProps helper or this rule to be more custom, but we should add an invalid test that checks whether an existing `variant` prop is replaced or not:

```javascript
{
  code:   `import { Label } from '@patternfly/react-core'; <Label isOverflowLabel variant="outline" />`,
  output: `import { Label } from '@patternfly/react-core'; <Label variant="overflow" />`,
  errors: [{
      message: `isOverflowLabel prop for Label has been replaced with variant="overflow"`,
      type: "JSXOpeningElement",
      }]
    },
```

I'd be fine having this be a followup, though, especially since we can't be sure if consumers would be passing both isOverflowLabel and a variant (since the variant shouldn't have any effect for an overflow label I believe).

packages/eslint-plugin-pf-codemods/src/rules/v6/labelRemoveIsOverflowLabel/label-remove-isOverflowLabel.md
```diff
@@ -0,0 +1,17 @@
+### label-remove-isOverflowLabel [(#10037)](https://github.com/patternfly/patternfly-react/pull/10037)
+
+isOverflowLabel prop for Label has been replaced with variant="overflow"
```
thatblindgeye, line 3: Related to above comment, we should add onto the message here + when an error is thrown: "Running the fix for this rule will replace an existing `variant` prop." or something along those lines.

thatblindgeye (approved):
Couple comments below, not a blocker for this PR but worth opening a followup issue for.

packages/eslint-plugin-pf-codemods/src/rules/v6/labelRemoveIsOverflowLabel/label-remove-isOverflowLabel.test.ts
adamviktora: Good point! I think we can do the change right now, I will make this rule more custom

thatblindgeye (approved):
lgtm. @wise-king-sullyman would you prefer having the open PRs updated to add typing, or do that after alpha when we might have more time to?

adamviktora:
I updated types on this and the Nav PR to use the types from eslint

packages/eslint-plugin-pf-codemods/src/rules/v6/labelRemoveIsOverflowLabel/label-remove-isOverflowLabel.ts
```diff
@@ -0,0 +1,64 @@
...
+              node.name.type === "JSXIdentifier" &&
+              labelImport.local.name === node.name.name
+            ) {
+              const isOverflowLabelAttribute = node.attributes.find(
+                (attr) =>
+                  attr.type === "JSXAttribute" &&
+                  attr.name?.name === "isOverflowLabel"
+              );
+
+              if (isOverflowLabelAttribute) {
```
wise-king-sullyman, line 30: Nit: I'd rather see this check setup as a guard clause with an early return than have the whole rest of the logic nested one level deeper. IMO that layout just keeps things a little bit simpler.

wise-king-sullyman:
@thatblindgeye ATM I consider typing just a nice to have. After we have the mods for alpha done I would move it up in priority yeah.

The types you've added so far look great though Adam! 🔥

wise-king-sullyman (approved):
One tiny non-blocking nit/comment, this looks great though 🚀

wise-king-sullyman (approved):
🔥

PR #581, opened 2024-02-20
feat(FormGroup): rename labelIcon to labelHelp

### LINKED ISSUE #563: FormGroup - Penta updates
Follow up to breaking change PR: https://github.com/patternfly/patternfly-react/pull/10016

Rename `labelIcon` prop to `labelHelp` in `FormGroup` component.

_Required actions:_
1. Build codemod
2. Build test

adamviktora:
Closes #563

Wait until the underlying react PR https://github.com/patternfly/patternfly-react/pull/10016 is approved and merged before merging this codemod PR.

Possible improvements / thoughts:
- trying to adjust whatever consumer is passing to `labelIcon` to use the new FormGroupLabelHelp component is nearly impossible and there would be lots of edge cases, so I would rather not do that
- can we add a comment directly to the code, which would recommend using `FormGroupLabelHelp` component for the `labelHelp` prop? Something like: 
input:
```
<FormGroup labelIcon={<button>Help icon</button>} />
```

output:
```
<FormGroup labelHelp={/* CODEMODS RECOMMENDATION: use FormGroupLabelHelp component */ <button>Help icon</button>} />
```
or is a message in the codemod error sufficient?

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/formGroupRenameLabelIcon/formGroup-rename-labelIcon.md
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/formGroupRenameLabelIcon/formGroup-rename-labelIcon.md
@@ -1,0 +1,18 @@
+### formGroup-rename-labelIcon [(#10016)](https://github.com/patternfly/patternfly-react/pull/10016)
+
+The `labelIcon` prop for FormGroup has been renamed to `labelHelp`. We recommend using FormGroupLabelHelp element for the labelHelp prop. The markup has also changed, we now wrap the labelHelp element in `<span className={styles.formGroupLabelHelp}>`, so there is no need to add `className={styles.formGroupLabelHelp}` to the labelHelp element.
+
+#### Examples
+
+In:
+
+```jsx
+%inputExample%
+```
+
+Out:
+
+```jsx
+%outputExample%
+```
+
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/formGroupRenameLabelIcon/formGroup-rename-labelIcon.md
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/formGroupRenameLabelIcon/formGroup-rename-labelIcon.md
@@ -15,4 +15,3 @@ Out:
 ```jsx
 %outputExample%
 ```
-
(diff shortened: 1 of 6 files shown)


thatblindgeye (commented):
One comment below. Agreed that we shouldn't try to update code to render FormGroupLabelHelp, that should be on the consumer once they're made aware of the update + our recommendation.

As for adding a comment inline with wherever the labelHelp prop is, I'd probably lean towards not doing that, since it's clutter being added to consumer code that may not be necessary. The consumer could already have their own component being passed that is essentially our FormGroupLabelHelp or something, and the error message will already recommend it.

packages/eslint-plugin-pf-codemods/src/rules/v6/formGroupRenameLabelIcon/formGroup-rename-labelIcon.test.ts
```diff
@@ -0,0 +1,24 @@
...
+      code:   `import { FormGroup } from '@patternfly/react-core'; <FormGroup labelIcon={<>Help icon</>} />`,
+      output: `import { FormGroup } from '@patternfly/react-core'; <FormGroup labelHelp={<>Help icon</>} />`,
+      errors: [{
+        message: `The \`labelIcon\` prop for FormGroup has been renamed to \`labelHelp\`. We recommend using FormGroupLabelHelp element for the labelHelp prop. The markup has also changed, we now wrap the labelHelp element in \`<span className={styles.formGroupLabelHelp}>\`, so there is no need to add \`className={styles.formGroupLabelHelp}\` to the labelHelp element.`,
+        type: "JSXOpeningElement",
+      }]
+    },
+  ]
+});
+
```
wise-king-sullyman, line 24: Nit: looks like this file could use a prettier run.

wise-king-sullyman (approved):
One nit and agree with @thatblindgeye's thoughts above, nothing blocking that I see though.

packages/eslint-plugin-pf-codemods/src/rules/v6/formGroupRenameLabelIcon/formGroup-rename-labelIcon.md
```diff
@@ -0,0 +1,18 @@
+### formGroup-rename-labelIcon [(#10016)](https://github.com/patternfly/patternfly-react/pull/10016)
+
+The `labelIcon` prop for FormGroup has been renamed to `labelHelp`. We recommend using FormGroupLabelHelp element for the labelHelp prop. The markup has also changed, we now wrap the labelHelp element in `<span className={styles.formGroupLabelHelp}>`, so there is no need to add `className={styles.formGroupLabelHelp}` to the labelHelp element.
```
thatblindgeye, line 3: I may have not added a comment when I left a review last week, but pretty sure it was for the message relayed to the consumer. Instead of styles.formGroupLabelHelp, feel like we should use the actual class string instead. Would be for here and the tests/actual rule code.

thatblindgeye (approved):

wise-king-sullyman (approved):

PR #588, opened 2024-02-26
feat(Chip) update to deprecated path

### LINKED ISSUE #557: Chip/ChipGroup - update to deprecated path
`Chip` and `ChipGroup` have been deprecated and need their path updated.

Dominik-Petrik:
close #557


wise-king-sullyman:
Looks like your invalid test is failing due to a line break difference

wise-king-sullyman (commented):
Other than the failing test looks great.

thatblindgeye (approved):
@wise-king-sullyman do you think we still need to add tests checking for esm imports etc similar to the v5 deprecated rules? Not for this PR but as a followup, this PR looks good to merge in as is to me.

wise-king-sullyman (approved):

Issue #541, opened 2023-10-30
table-rename-isHoverable: Cannot read properties of undefined (reading 'name')


Andrei-Stepanov:
Hello,

Could you please help me?
I am following: https://www.patternfly.org/get-started/upgrade/release-notes/
I get an error:

```
➜  osci npx @patternfly/pf-codemods@latest ciboard
Need to install the following packages:
  @patternfly/pf-codemods@1.82.1
Ok to proceed? (y) 
/home/astepano/.npm/_npx/d72ba669e7bba25a/node_modules/@patternfly/eslint-plugin-pf-codemods/lib/rules/v5/table-rename-isHoverable.js:10
    ).imports.filter((specifier) => ["Table", "Tr"].includes(specifier.imported.name));
                                                                                ^

TypeError: Error while loading rule '@patternfly/pf-codemods/table-rename-isHoverable': Cannot read properties of undefined (reading 'name')
Occurred while linting /home/astepano/osci/ciboard/src/components/SSTResultsTable.tsx
    at /home/astepano/.npm/_npx/d72ba669e7bba25a/node_modules/@patternfly/eslint-plugin-pf-codemods/lib/rules/v5/table-rename-isHoverable.js:10:81
    at Array.filter (<anonymous>)
    at Object.create (/home/astepano/.npm/_npx/d72ba669e7bba25a/node_modules/@patternfly/eslint-plugin-pf-codemods/lib/rules/v5/table-rename-isHoverable.js:10:15)
    at createRuleListeners (/home/astepano/.npm/_npx/d72ba669e7bba25a/node_modules/eslint/lib/linter/linter.js:910:21)
    at /home/astepano/.npm/_npx/d72ba669e7bba25a/node_modules/eslint/lib/linter/linter.js:1081:110
    at Array.forEach (<anonymous>)
    at runRules (/home/astepano/.npm/_npx/d72ba669e7bba25a/node_modules/eslint/lib/linter/linter.js:1018:34)
    at Linter._verifyWithoutProcessors (/home/astepano/.npm/_npx/d72ba669e7bba25a/node_modules/eslint/lib/linter/linter.js:1370:31)
    at Linter._verifyWithConfigArray (/home/astepano/.npm/_npx/d72ba669e7bba25a/node_modules/eslint/lib/linter/linter.js:1822:21)
    at Linter.verify (/home/astepano/.npm/_npx/d72ba669e7bba25a/node_modules/eslint/lib/linter/linter.js:1452:65)
[...]

PR #590, opened 2024-02-26
feat(Menu): updated markup for MenuItemAction

### LINKED ISSUE #589: Menu - updated markup on MenuItemAction
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/10089

Warn consumers that the markup has changed

_Required actions:_
1. Build codemod
2. Build test

thatblindgeye:
Closes #589

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/menuItemActionWarnUpdateMarkup/menuItemAction-warn-update-markup.ts
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/menuItemActionWarnUpdateMarkup/menuItemAction-warn-update-markup.ts
@@ -2,17 +2,17 @@ import { getFromPackage } from "../../helpers";
 import { Rule } from "eslint";
 import { ImportDeclaration } from "estree-jsx";
 
-// https://github.com/patternfly/patternfly-react/pull/10027
+// https://github.com/patternfly/patternfly-react/pull/10089
 module.exports = {
   meta: {},
   create: function (context: Rule.RuleContext) {
     const { imports } = getFromPackage(context, "@patternfly/react-core");
 
-    const jumpLinksItemImport = imports.find(
-      (specifier) => specifier.imported.name === "JumpLinksItem"
+    const menuItemActionImport = imports.find(
+      (specifier) => specifier.imported.name === "MenuItemAction"
     );
 
-    return !jumpLinksItemImport
+    return !menuItemActionImport
       ? {}
       : {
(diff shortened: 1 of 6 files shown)


packages/eslint-plugin-pf-codemods/src/rules/v6/menuItemActionWarnUpdateMarkup/menuItemAction-warn-update-markup.ts
```diff
@@ -0,0 +1,35 @@
...
+    );
+
+    return !menuItemActionImport
+      ? {}
+      : {
+          ImportDeclaration(node: ImportDeclaration) {
+            if (
+              node.specifiers.find(
+                (specifier) =>
+                  specifier.type === "ImportSpecifier" &&
```
adamviktora, line 22: I was thinking about what if someone imports Patternfly and uses it this way:

```
import * as PF from '@patternfly/react-core';

<PF.MenuItemAction></PF.MenuItemAction>
```
I hope nobody does it, I checked the `getFromPackage` helper and we don't handle this `ImportNamespaceSpecifier` there either. So if we don't know of any team who would do it, we don't have to worry about it, it would bring too many complications.

adamviktora (approved):
LPTM (looks perfect to me)! 

(one comment bellow, but probably not something we should worry about)

packages/eslint-plugin-pf-codemods/src/rules/v6/menuItemActionWarnUpdateMarkup/menuItemAction-warn-update-markup.ts
thatblindgeye: Ah that's an interesting point! I don't think we had any consumers bring this to our attention during the v5 release (@wise-king-sullyman do you recall if we did?), but worth keeping an eye out for with v6 alpha and beta releases

wise-king-sullyman: That's in the same realm as the root problem behind #541, IMO it is probably worth us properly handling all of those import cases, but I don't think ironing those edge cases out is worth holding up the alpha release of the mods.

I think TS will be a big help with these kinds of problems, typing `getFromPackage` made it pretty obvious that we aren't logically setup for all of the potential import declarations.

wise-king-sullyman (approved):

wise-king-sullyman (approved):

PR #591, opened 2024-02-26
feat(Card): Remove deprecated and unused props

### LINKED ISSUE #560: [Card] - Remove deprecated props
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/10056

Removed the following props that were previosly deprecated and are no longer supported in Penta:
- isSelectableRaised
- isDisabledRaised
- hasSelectableInput
- selectableInputAriaLabel
- onSelectableInputChange

Also removed the following props for Penta:
- isFlat
- isRounded

_Required actions:_
1. Build codemod
2. Build test

wise-king-sullyman:
Closes #560

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/cardRemoveVariousProps/card-remove-various-props.md
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/cardRemoveVariousProps/card-remove-various-props.md
@@ -1,0 +1,18 @@
+### card-remove-various-props [(#10056)](https://github.com/patternfly/patternfly-react/pull/10056)
+
+The isSelectableRaised, isDisabledRaised, hasSelectableInput, selectableInputAriaLabel, onSelectableInputChange, isFlat, and isRounded props have been removed
+
+#### Examples
+
+In:
+
+```jsx
+%inputExample%
+```
+
+Out:
+
+```jsx
+%outputExample%
+```
+
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/cardRemoveVariousProps/card-remove-various-props.md
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/cardRemoveVariousProps/card-remove-various-props.md
@@ -1,6 +1,13 @@
 ### card-remove-various-props [(#10056)](https://github.com/patternfly/patternfly-react/pull/10056)
 
-The isSelectableRaised, isDisabledRaised, hasSelectableInput, selectableInputAriaLabel, onSelectableInputChange, isFlat, and isRounded props have been removed
+The following props have been removed from Card:
+- isSelectableRaised
+- isDisabledRaised
+- hasSelectableInput
+- selectableInputAriaLabel
+- onSelectableInputChange
+- isFlat
+- isRounded
 
 #### Examples
 
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/cardRemoveVariousProps/card-remove-various-props.test.ts
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/cardRemoveVariousProps/card-remove-various-props.test.ts
@@ -1,0 +1,77 @@
+const ruleTester = require("../../ruletester");
+import * as rule from "./card-remove-various-props";
+
+interface InvalidTest {
+  code: string;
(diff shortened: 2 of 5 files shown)


packages/eslint-plugin-pf-codemods/src/rules/v6/cardRemoveVariousProps/card-remove-various-props.md
```diff
@@ -0,0 +1,18 @@
+### card-remove-various-props [(#10056)](https://github.com/patternfly/patternfly-react/pull/10056)
+
+The isSelectableRaised, isDisabledRaised, hasSelectableInput, selectableInputAriaLabel, onSelectableInputChange, isFlat, and isRounded props have been removed
```
thatblindgeye, line 3: These are more nits so I'll leave it up to you what you prefer:

- adding "from Card" at the end to match how we've generally wrote these messages (even though the rule name states it and it'll be flagged on the applicable component when codemods are ran)
- an unordered list of the props instead of comma separated, e.g. `The following props have been removed from Card: ...`

packages/eslint-plugin-pf-codemods/src/rules/v6/cardRemoveVariousProps/card-remove-various-props.test.ts
```diff
@@ -0,0 +1,77 @@
...
+      code: `import { Card } from '@patternfly/react-core'; <Card ${prop}='foo bar' />`,
+      output: `import { Card } from '@patternfly/react-core'; <Card  />`,
+      errors: [{ message: `The ${prop} prop has been removed` }],
+    },
+    {
+      code: `import { Card } from '@patternfly/react-core'; const foo = "foo bar"; <Card ${prop}={foo} />`,
+      output: `import { Card } from '@patternfly/react-core'; const foo = "foo bar"; <Card  />`,
+      errors: [{ message: `The ${prop} prop has been removed` }],
+    },
+  ];
```
thatblindgeye, line 39: Not opposed to including tests like this for rename/removal of props, just curious if you foresee not testing for it being an issue? Since the helper itself doesn't really care what's passed into the prop, just its existence.

packages/eslint-plugin-pf-codemods/src/rules/v6/cardRemoveVariousProps/cardRemoveVariousPropsOutput.tsx
```diff
@@ -0,0 +1,3 @@
+import { Card } from "@patternfly/react-core";
+
+export const CardRemoveVariousPropsInput = () => <Card />;
```
thatblindgeye, line 3: This output should include the whitespace left behind by the removed props:

```
  <Card
    
    
    
    
    
  />
```

thatblindgeye (commented):
Mostly nits below, one possible change to the Output file

packages/eslint-plugin-pf-codemods/src/rules/v6/cardRemoveVariousProps/card-remove-various-props.test.ts
wise-king-sullyman: Sooooo long term I actually think we should change this by having dedicated tests for the helpers themselves. Until we can do that though I feel like these additional tests are worth it on the actual rule side.

packages/eslint-plugin-pf-codemods/src/rules/v6/cardRemoveVariousProps/cardRemoveVariousPropsOutput.tsx
wise-king-sullyman: Whoops, blame prettier 😆

packages/eslint-plugin-pf-codemods/src/rules/v6/cardRemoveVariousProps/card-remove-various-props.test.ts
thatblindgeye: Yeah I was thinking that as well  re: tests for helpers. I'm good with keeping these tests, though, as it doesn't really hurt.

thatblindgeye (approved):

PR #602, opened 2024-03-05
refactor(EmptyState): extract helpers, fix(EmptyState): getElementChildText fix

### LINKED ISSUE #594: Helpers - Extract & refine small helpers from #582
Follow up to PR #582

Many of the small helper utilities introduced there would likely be useful in other rules, they should be extracted and refined where needed.

adamviktora:
Closes #594 

Also applies a fix to a function which gets children text of EmptyStateHeader.

Previously, situations like this:
```
<EmptyStateHeader>
  <h1>Title</h1>
</EmptyStateHeader>
```
were caught in switch command under `JSXText` (because of the initial whitespaces before `<h1>`).
And if not, it ended up always being wrapped in `""`, which we don't want, if the children is not a single `JSXText` element.


packages/eslint-plugin-pf-codemods/src/rules/v6/emptyStateHeaderMoveIntoEmptyState/emptyStateHeader-move-into-emptyState.ts
```diff
@@ -171,21 +78,56 @@ module.exports = {
...
+                singleChild as JSXExpressionContainer | JSXElement | JSXFragment
+              );
+
+            return singleChild.type === "JSXExpressionContainer"
+              ? singleChildText
+              : `{${singleChildText}}`;
+          }
+
+          return `{<>${getNodesText(context, children as Node[])}</>}`;
+        };
```
wise-king-sullyman, line 125: Would it not make sense to extract this to your getText file?

packages/eslint-plugin-pf-codemods/src/rules/v6/emptyStateHeaderMoveIntoEmptyState/emptyStateHeader-move-into-emptyState.test.ts
```diff
@@ -148,6 +148,112 @@ ruleTester.run("emptyStateHeader-move-into-emptyState", rule, {
...
+          </EmptyState>
+      );
+      `,
+      errors: [
+        {
+          message: `EmptyStateHeader has been moved inside of the EmptyState component and is now only customizable using props, and the EmptyStateIcon component now wraps content passed to the icon prop automatically. Additionally, the titleText prop is now required on EmptyState.`,
+          type: "JSXElement",
+        },
+      ],
+    },
```
wise-king-sullyman, line 256: Great catch finding these problems, fixing them, and adding tests 💪

wise-king-sullyman (commented):
Looking great 🔥 

One question/suggestion then it should be good to go.

packages/eslint-plugin-pf-codemods/src/rules/v6/emptyStateHeaderMoveIntoEmptyState/emptyStateHeader-move-into-emptyState.ts
adamviktora: Probably yes, I at first thought it is too specific use case, but as I am looking at it, it can be extracted. I renamed it to `getChildrenAsAttributeValueText`. This name makes more sense to me, because it is transforming children to a text which is being used as an attribute value (including the expression container or quotation marks).

wise-king-sullyman (approved):
🚀

packages/eslint-plugin-pf-codemods/src/rules/helpers/JSXAttributes.ts
```diff
@@ -0,0 +1,18 @@
+import { JSXElement, JSXAttribute } from "estree-jsx";
+
+export function getAttribute(node: JSXElement, attributeName: string) {
+  const attributes = node.openingElement.attributes.filter(
+    (attr) => attr.type === "JSXAttribute"
+  ) as JSXAttribute[];
+  return attributes.find((attr) => attr.name.name === attributeName);
```
Dominik-Petrik, line 7: Just a nit, is there a benefit in filtering the attributes first by type and then finding the one with correct name as opposed to just using single `find()`
```suggestion
  return node.openingElement.attributes.find((attr) => attr.type === "JSXAttribute" && attr.name.name === attributeName);
```

Dominik-Petrik (commented):
Other than Austin comments and the nit LGTM! 🚀

packages/eslint-plugin-pf-codemods/src/rules/helpers/JSXAttributes.ts
adamviktora: Good catch! Thanks for the review

adamviktora: oh, wait, something is failing

Dominik-Petrik (approved):

Issue #573, opened 2024-02-19
Type renameProps and findAncestor helpers


thatblindgeye:
Part of the effort to type our helpers

_Required actions:_
1. Build codemod
2. Build test
3. Update readme with description & example

PR #603, opened 2024-03-05
chore(helpers): extract findAncestor helper

thatblindgeye:
Towards #573

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/rules/helpers/findAncestor.ts
+++ b/packages/eslint-plugin-pf-codemods/src/rules/helpers/findAncestor.ts
@@ -22,32 +22,3 @@ export function findAncestor(
     current = current.parent;
   }
 }
-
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin-pf-codemods/src/rules/helpers/findAncestor.ts
```diff
@@ -0,0 +1,28 @@
+import { JSXElement, JSXOpeningElement } from "estree-jsx";
+
+type ElementWithParent = (JSXOpeningElement | JSXElement) & {
+  parent?: ElementWithParent | null;
+};
+
+export function findElementAncestor(
+  node: ElementWithParent,
+  ancestorNameToFind: string
```
thatblindgeye, line 9: Since this helper has only been used to find an element ancestor, tweaked this from the original helper. Kept the file name generic in case we want to add other helpers, like a general "findNodeAncestor" or something.

wise-king-sullyman (approved):

PR #604, opened 2024-03-07
chore(deps): upgrade TypeScript to latest version

jonkoops:
Upgrades TypeScript to the latest available version. This helps projects that are trying to apply the codemod on modern TypeScript code bases, as those will currently error out on unsupported syntax.


mfrances17 (approved):
LGTM 👍

mfrances17:
@wise-king-sullyman @thatblindgeye @tlabaj if we update the typescript version, will this be a problem for products running the codemods who are using earlier TS versions? it will require that they update the TS version for their product in order to run the codemods. updating TS could _potentially_ introduce issues in their product and be considered a 'breaking change' of sorts.

wise-king-sullyman (approved):
Thanks!

PR #607, opened 2024-03-11
chore(devex) Add types for renameProps helper

Dominik-Petrik:
Towards #573


packages/eslint-plugin-pf-codemods/src/rules/helpers/renamePropsOnNode.ts
```diff
@@ -0,0 +1,108 @@
...
+              `${attribute.name.name} prop for ${
+                (node.name as JSXIdentifier).name
+              } has been ${
+                newPropObject.replace ? 'replaced with' : 'renamed to'
+              } ${newName}`,
+            fix(fixer) {
+              return fixer.replaceText(
+                newPropObject.replace
+                  ? attribute
+                  : (attribute as JSXAttribute).name,
```
wise-king-sullyman, line 88: I'm curious if this assertion is actually definitive here as well.

packages/eslint-plugin-pf-codemods/src/rules/helpers/renamePropsOnNode.ts
```diff
@@ -0,0 +1,108 @@
...
+            },
+          });
+        } else {
+          context.report({
+            node,
+            message:
+              message ||
+              `${attribute.name.name} prop for ${
+                (node.name as JSXIdentifier).name
+              } has been removed`,
```
wise-king-sullyman, line 100: Same comments as above regarding the message and the type assertion.

wise-king-sullyman (changes requested):
Awesome work 🚀 

I've got a good amount of comments here, but many are just general refactor things that I'm absolutely fine with being a followup.

packages/eslint-plugin-pf-codemods/src/rules/helpers/renameSinglePropOnNode.ts
```diff
@@ -0,0 +1,99 @@
...
+  if (removed) return 'removed';
+  return propRename.replace ? 'replaced with ' : 'renamed to ';
+};
+
+const getMessage = (
+  node: JSXOpeningElement,
+  propRename: RenameConfig | string | Record<PropertyKey, never>,
+  defaultMessage: string
+) => {
+  if (!(isRenameConfig(propRename) && propRename.message))
```
wise-king-sullyman, line 47: Nit: IMO it's really easy to misread a notted multi conditional like this and think the logic is actually `!isRenameConfig(propRename) && propRename.message`, I would not both expressions.

```suggestion
  if (!isRenameConfig(propRename) && !propRename.message)
```

packages/eslint-plugin-pf-codemods/src/rules/helpers/renameSinglePropOnNode.ts
```diff
@@ -0,0 +1,99 @@
...
+  propRename: RenameConfig | string | Record<PropertyKey, never>
+) {
+  if (attribute.name.type !== 'JSXIdentifier') return;
+
+  const isConfig = isRenameConfig(propRename);
+
+  const isRemoved = !(
+    (isConfig && propRename.newName !== '') ||
+    (typeof propRename === 'string' && propRename !== '')
+  );
```
wise-king-sullyman, line 68: Similar comment here, I would just refactor this to something like 
```suggestion
  const isReplaced = (isConfig && propRename.newName !== '') || (typeof propRename === 'string' && propRename !== '');
```

packages/eslint-plugin-pf-codemods/src/rules/helpers/renameSinglePropOnNode.ts
```diff
@@ -0,0 +1,99 @@
...
+    });
+  } else {
+    context.report({
+      node,
+      message,
+      fix(fixer) {
+        return fixer.replaceText(attribute, '');
+      },
+    });
+  }
```
wise-king-sullyman, line 98: Nit: I would refactor this to only call `context.report` once with a declared `fix` function that is customized via logic so that we have less duplicate code.

packages/eslint-plugin-pf-codemods/src/rules/helpers/renameSinglePropOnNode.ts
```diff
@@ -0,0 +1,99 @@
...
+  } has been ${action}${newName}`;
+  const message = getMessage(node, propRename, defaultMessage);
+
+  if (!isRemoved) {
+    context.report({
+      node,
+      message,
+      fix(fixer) {
+        return fixer.replaceText(
+          propRename.replace ? attribute : (attribute as JSXAttribute).name,
```
wise-king-sullyman, line 85: Also I think this type assertion can be removed now since attribute is typed as a `JSXAttribute`

wise-king-sullyman (commented):
This is looking GREAT 🔥 🔥 🔥, just a few more small things

packages/eslint-plugin-pf-codemods/src/rules/helpers/renameSinglePropOnNode.ts
Dominik-Petrik: I might be wrong but these two statements are not equal, are they? 🤔

Dominik-Petrik: And `Record<PropertyKey, never>` should be a [correct way](https://www.totaltypescript.com/the-empty-object-type-in-typescript) to type an empty object `{}`, which can be passed as a shorthand for removing a prop.

packages/eslint-plugin-pf-codemods/src/rules/helpers/renameSinglePropOnNode.ts
```diff
@@ -0,0 +1,94 @@
+import { JSXAttribute, JSXIdentifier, JSXOpeningElement } from 'estree-jsx';
+import { RenameConfig } from './renamePropsOnNode';
```
thatblindgeye, line 2: Rather than renameSingle importing something from renamePropOnNode, and in turn renamePropOnNode having to import from renameSingle, could we define the RenameConfig in this file then just import both into renamePropOnNode?

packages/eslint-plugin-pf-codemods/src/rules/helpers/renameSinglePropOnNode.ts
```diff
@@ -0,0 +1,94 @@
...
+import { Rule } from 'eslint';
+
+function isRenameConfig(obj: any): obj is RenameConfig {
```
thatblindgeye, line 5: May not be a blocker, but rather than typing this as any, wdyt about tweaking this function a little so that the param type matches what it would be passed in (`RenameConfig | string | Record<PropertyKey, never>`), changing the param name (since it won't only be an obj that can be passed), then adding an early return of false if `!hasValidObjectType`?

Also can we update the function name to something like `checkIsRenameConfig`

packages/eslint-plugin-pf-codemods/src/rules/helpers/renameSinglePropOnNode.ts
```diff
@@ -0,0 +1,94 @@
...
+    isValidMessage &&
+    isValidReplace
+  );
+}
+
+const getNewName = (
+  replaced: boolean,
+  propRename: RenameConfig | string | Record<PropertyKey, never>
+) => {
+  if (!replaced) return '';
```
thatblindgeye, line 30: This is partially a nit/personal opinion and partially for consistency, but rather than a single line return for the if blocks in these updates maybe we include the curly braces and put the return on a new line.

Not a huge thing, so if nobody else feels strongly about changing this then I'd be fine with it as-is.

packages/eslint-plugin-pf-codemods/src/rules/helpers/renameSinglePropOnNode.ts
```diff
@@ -0,0 +1,94 @@
...
+  return (
+    isRenameConfig(propRename) ? propRename.newName : propRename
+  ) as string;
```
thatblindgeye, line 33: Rather than having to assert here, wdyt about updating the if statement above to check whether `!replace || (![is not a rename config] && [is not a string])`, so that if propRename is an empty object we return the empty string as well?

thatblindgeye (approved):

packages/eslint-plugin-pf-codemods/src/rules/helpers/renameSinglePropOnNode.ts
wise-king-sullyman: Shoot I totally missed these messages, my bad. And yep you're right about the logic!

For the Record type TIL about empty object types, thanks for linking that! I do question the API a bit here, using an empty object for prop removal feels a bit odd for me, but that's something we can discuss later.

wise-king-sullyman (approved):
🚀 🚀 🚀 B E A U T I F U L 🚀 🚀 🚀

PR #608, opened 2024-03-12
feat(LoginMainFooterLinksItem): pass children instead of props

### LINKED ISSUE #598: Login - LoginMainFooterLinksItem structure update
Follow up to https://github.com/patternfly/patternfly-react/pull/10107

Instead of passing properties to the component, the content will be passed directly via `children`. There should be enough info from the props to generate the children (if the component was "a" create a link or if it was a "button" use our plain variant Button) but I'm wondering if that is still making too many assumptions.

adamviktora:
Closes #598

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/loginMainFooterLinksItemStructureUpdated/loginMainFooterLinksItem-structure-updated.test.ts
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/loginMainFooterLinksItemStructureUpdated/loginMainFooterLinksItem-structure-updated.test.ts
@@ -92,6 +92,18 @@ ruleTester.run("loginMainFooterLinksItem-structure-updated", rule, {
       </Button></LoginMainFooterLinksItem>`,
       errors: [loginMainFooterLinksItemError],
     },
+    // without attributes
+    {
+      code: `import { LoginMainFooterLinksItem } from '@patternfly/react-core';
+      <LoginMainFooterLinksItem>
+        <i>GitHub icon</i>
+      </LoginMainFooterLinksItem>`,
+      output: `import { LoginMainFooterLinksItem } from '@patternfly/react-core';
+      <LoginMainFooterLinksItem data-codemods="true" ><Button variant="link" component="a" href="">
+        <i>GitHub icon</i>
+      </Button></LoginMainFooterLinksItem>`,
+      errors: [loginMainFooterLinksItemError],
+    },
     // self-closing with children as attribute
     {
       code: `import { LoginMainFooterLinksItem } from '@patternfly/react-core';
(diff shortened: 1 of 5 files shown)


packages/eslint-plugin-pf-codemods/src/rules/v6/loginMainFooterLinksItemStructureUpdated/loginMainFooterLinksItem-structure-updated.test.ts
```diff
@@ -0,0 +1,128 @@
...
+  message: `LoginMainFooterLinksItem structure has changed. Instead of passing it many properties, everything is now passed directly in a children component.`,
+  type: "JSXElement",
+};
+
+ruleTester.run("loginMainFooterLinksItem-structure-updated", rule, {
+  valid: [
+    {
+      code: `<LoginMainFooterLinksItem />`,
+    },
+  ],
```
wise-king-sullyman, line 14: Could we add some additional valid tests here, for instance when it is a PF imported component but it's being used without the props that have been removed.

wise-king-sullyman (commented):
Looking great! I just have one thought with regards to the tests.

packages/eslint-plugin-pf-codemods/src/rules/v6/loginMainFooterLinksItemStructureUpdated/loginMainFooterLinksItem-structure-updated.test.ts
adamviktora: I think we have to apply a fix in every case it is being imported from PF. There will always be a default `linkComponent` attribute (`a`). So we always have to pass a children wrapped in the component which was internal previously (https://github.com/patternfly/patternfly-react/pull/10107/files)

wise-king-sullyman (approved):

thatblindgeye (changes requested):
Overall I think this looks good. One issue is that while we're updating the component to pass the proper children (with its own props), we're not checking whether the Button component is imported and adding it to the imports if not. Since the fix requires the Button import, I think it'd make sense to include the checking (and importing if not found) of the Button component.

packages/eslint-plugin-pf-codemods/src/rules/v6/loginMainFooterLinksItemStructureUpdated/loginMainFooterLinksItem-structure-updated.test.ts
```diff
@@ -0,0 +1,164 @@
...
+      ></LoginMainFooterLinksItem>`,
+      output: `import { LoginMainFooterLinksItem, Button } from '@patternfly/react-core';
+      <LoginMainFooterLinksItem data-codemods="true" 
+        
+        
+        
+      ><Button variant="link" component="a" href="https://github.com/login" {...{ "aria-label": "Login with Github" }}><i>GitHub icon</i></Button></LoginMainFooterLinksItem>`,
+      errors: [loginMainFooterLinksItemError],
+    },
+    // with Button component already imported
```
thatblindgeye, line 139: 🔥

packages/eslint-plugin-pf-codemods/src/rules/v6/loginMainFooterLinksItemStructureUpdated/loginMainFooterLinksItemStructureUpdatedOutput.tsx
```diff
@@ -0,0 +1,14 @@
+import { LoginMainFooterLinksItem } from "@patternfly/react-core";
```
thatblindgeye, line 1: The fix command just needs to be ran to update the output so that Button gets imported

thatblindgeye (changes requested):

thatblindgeye (approved):

PR #610, opened 2024-03-12
feat(FormFiledGroupHeaderTitleTextObject): interface renamed (typo)

### LINKED ISSUE #609: FormFiledGroupHeaderTitleTextObject - typo, rename to FormFieldGroupHeaderTitleTextObject
Follow up to breaking change [PR](https://github.com/patternfly/patternfly-react/pull/10016)

There was a typo in exported interface: Form**Filed**GroupHeaderTitleTextObject, rename it to FormFieldGroupHeaderTitleTextObject.

_Required actions:_
1. Build codemod
2. Build test

adamviktora:
Closes #609

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/formFiledGroupHeaderTitleTextObjectRenamed/formFiledGroupHeaderTitleTextObject-renamed.ts
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/formFiledGroupHeaderTitleTextObjectRenamed/formFiledGroupHeaderTitleTextObject-renamed.ts
@@ -1,0 +1,59 @@
+import { AST, Rule } from "eslint";
+import { ImportSpecifier, Identifier, Node } from "estree-jsx";
+import { getFromPackage } from "../../helpers";
+
+// https://github.com/patternfly/patternfly-react/pull/10016
+module.exports = {
+  meta: { fixable: "code" },
+  create: function (context: Rule.RuleContext) {
+    const { imports } = getFromPackage(context, "@patternfly/react-core");
+
+    const previousName = "FormFiledGroupHeaderTitleTextObject";
+    const newName = "FormFieldGroupHeaderTitleTextObject";
+
+    const interfaceImport = imports.find(
+      (specifier) => specifier.imported.name === previousName
+    );
+
+    if (!interfaceImport) {
+      return {};
+    }
+
+    const hasAlias = (specifier: ImportSpecifier) =>
+      specifier.local.name !== specifier.imported.name;
+
+    const reportMessage =
+      "There was a typo in FormFiledGroupHeaderTitleTextObject interface. It was renamed to the intended FormFieldGroupHeaderTitleTextObject.";
+
+    const callContextReport = (node: Node, nodeToReplace: Node | AST.Token) => {
+      context.report({
+        node,
+        message: reportMessage,
+        fix(fixer) {
+          return fixer.replaceText(nodeToReplace, newName);
+        },
+      });
+    };
+
+    return {
+      ImportSpecifier(node: ImportSpecifier) {
+        if (node.imported.name === interfaceImport.imported.name) {
+          callContextReport(node, node.imported);
+        }
+      },
+      TSTypeReference(node: { typeName: Identifier }) {
+        if (!hasAlias(interfaceImport) && node.typeName.name === previousName) {
+          callContextReport(node as unknown as Node, node.typeName);
+        }
+      },
+      TSInterfaceHeritage(node: { expression: Identifier }) {
+        if (
+          !hasAlias(interfaceImport) &&
+          node.expression.name === previousName
+        ) {
+          callContextReport(node as unknown as Node, node.expression);
+        }
+      },
+    };
+  },
+};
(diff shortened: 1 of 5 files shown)


packages/eslint-plugin-pf-codemods/src/rules/v6/formFiledGroupHeaderTitleTextObjectRenamed/formFiledGroupHeaderTitleTextObject-renamed.ts
```diff
@@ -0,0 +1,59 @@
...
+        if (node.imported.name === interfaceImport.imported.name) {
+          callContextReport(node, node.imported);
+        }
+      },
+      TSTypeReference(node: { typeName: Identifier }) {
+        if (!hasAlias(interfaceImport) && node.typeName.name === previousName) {
+          callContextReport(node as unknown as Node, node.typeName);
+        }
+      },
+      TSInterfaceHeritage(node: { expression: Identifier }) {
```
wise-king-sullyman, line 49: Having these typed as objects with just the expected values and then asserting that they're Nodes feels odd, are there not better types we could use here?

wise-king-sullyman (commented):
Looks great! One type related question.

packages/eslint-plugin-pf-codemods/src/rules/v6/formFiledGroupHeaderTitleTextObjectRenamed/formFiledGroupHeaderTitleTextObject-renamed.ts
adamviktora: Yeah, I know. I was struggling to find any type for these, there are only enums for these in `/packages/eslint-plugin-pf-codemods/lib/rules/ast-node-types.d.ts` which is I think is copied from `@typescript-eslint/parser`. But I am taking a look at the typescript-eslint repo now and it seems the actual interface might be exported there.

I'll try to import that

wise-king-sullyman: Hm 🤔 

Let me do a little hunting myself and see if I can find the types we need exported anywhere.

wise-king-sullyman: I think I found them:

`import { TSTypeReference, TSInterfaceHeritage} from '@typescript-eslint/parser'`

adamviktora: I tried importing this, the '@typescript-eslint/parser' package should be installed, but it says it is not exported.

Screenshot 2024-03-25 at 11 27 55

Updated `packages/eslint-plugin-pf-codemods/package.json`:

Screenshot 2024-03-25 at 11 27 04

Then I did `yarn install`
I don't know if I am installing something wrong, but it just seems like it is not exported.

wise-king-sullyman: I'm seeing the same thing but I swear this was working on Friday 🤔

wise-king-sullyman: I really am sure that this was working Friday and now I'm so confused 😆 

It looks like types of those names can be retrieved from `import { TSESTree } from '@typescript-eslint/types'
`... but they don't slide right into place and I'm not sure if those are the wrong type or if the current logic isn't fully type safe.

I do feel like we should investigate this further but I won't block this PR over it if you'd like to make a followup issue so that we don't forget about it.

adamviktora: Typescript contributors are playing some high-level games with us I guess 😃 

Anyways, the `TSESTree.AST_NODE_TYPES` enum si fine, but it does not fit the use case as you say. I'd also not block over, the codemod should work even if typed as `TSTypeReference(node: any) {`, as if the node matches that TSTypeReference function, it should already have the suitable structure I suppose.

Opened the issue for further investigation: https://github.com/patternfly/pf-codemods/issues/622

wise-king-sullyman: Yeah the codemod will still work, which is why I'm fine with the typing being patched in a followup. It just isn't ideal from a typing perspective since it isn't _really_ valid, we're just telling TS to trust us and chill out 😆

wise-king-sullyman (approved):
🎉

thatblindgeye (approved):
Looks good, opened https://github.com/patternfly/pf-codemods/issues/623 to possibly add logic to update an instance where the object is being exported as well.

PR #612, opened 2024-03-14
feat(EmptyState): support icons passed directly to EmptyStateHeader

### LINKED ISSUE #595: EmptyState - support icons passed directly to EmptyStateHeader
Follow up to PR #582 

The current implementation doesn't support handling icons which were passed directly to the EmptyStateHeader rather than being wrapped in an EmptyStateIcon container, this issue should address that potential issue.

adamviktora:
Uses helpers from #602 which should be merged first before this PR.

Closes #595

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/emptyStateHeaderMoveIntoEmptyState/emptyStateHeader-move-into-emptyState.ts
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/emptyStateHeaderMoveIntoEmptyState/emptyStateHeader-move-into-emptyState.ts
@@ -157,13 +157,21 @@ module.exports = {
           });
         }
 
-        const icon = emptyStateIconComponentIconAttribute
-          ? context
+        const getIconPropText = () => {
+          if (emptyStateIconComponentIconAttribute) {
+            return context
               .getSourceCode()
-              .getText(emptyStateIconComponentIconAttribute)
-          : iconPropIsIconElement()
-          ? `icon={${iconElementIdentifier!.name}}`
-          : "";
+              .getText(emptyStateIconComponentIconAttribute);
+          }
+
+          if (iconPropIsIconElement()) {
+            return `icon={${iconElementIdentifier!.name}}`;
+          }
+
+          return "";
+        };
+
+        const icon = getIconPropText();
 
         context.report({
           node,
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/emptyStateHeaderMoveIntoEmptyState/emptyStateHeader-move-into-emptyState.ts
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/emptyStateHeaderMoveIntoEmptyState/emptyStateHeader-move-into-emptyState.ts
@@ -5,6 +5,7 @@ import {
   getAttributeText,
   getAttributeValueText,
   getChildElementByName,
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin-pf-codemods/src/rules/v6/emptyStateHeaderMoveIntoEmptyState/emptyStateHeader-move-into-emptyState.ts
```diff
@@ -129,8 +130,48 @@ module.exports = {
...
+          return (
+            iconElementIdentifier !== undefined &&
+            allIconSpecifiers.some(
+              (spec) => spec.local.name === iconElementIdentifier.name
+            )
+          );
+        };
+
+        const emptyStateIconComponent = iconPropIsEmptyStateIconComponent()
+          ? (iconPropValue as JSXElement)
```
wise-king-sullyman, line 173: Is this type assertion definitely valid?

packages/eslint-plugin-pf-codemods/src/rules/v6/emptyStateHeaderMoveIntoEmptyState/emptyStateHeader-move-into-emptyState.ts
```diff
@@ -155,6 +192,8 @@ module.exports = {
           ? context
               .getSourceCode()
               .getText(emptyStateIconComponentIconAttribute)
+          : iconPropIsIconElement()
+          ? `icon={${iconElementIdentifier!.name}}`
```
wise-king-sullyman, line 196: Is there a way to refactor this logic to avoid the nested ternary?

wise-king-sullyman (commented):
Lookin 🔥!

I just have one type question and a small refactor request.

packages/eslint-plugin-pf-codemods/src/rules/v6/emptyStateHeaderMoveIntoEmptyState/emptyStateHeader-move-into-emptyState.ts
adamviktora: Yes, it is quite indirect, but the `iconPropIsEmptyStateIconComponent` function also checks that `iconElementIdentifier` is not undefined, which is true when `iconPropValue?.type === "JSXElement"`.

wise-king-sullyman: What if there is no `EmptyStateIcon` import? Would that cause `emptyStateIconImport` to be `undefined` and `iconPropIsEmptyStateIconComponent` to incorrectly return true (since `undefined === undefined` is true)?

adamviktora: Oh, good catch, that could cause some trouble, thanks for finding that!

wise-king-sullyman: That's what I'm here for! 🙂

wise-king-sullyman (approved):

PR #617, opened 2024-03-19
feat(TreeView): warning for removal of selectable styling modifier

### LINKED ISSUE #601: TreeView - add warning for removal of selectable styling modifier
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/10101

We should add a warning regarding the removal of this modifier class.

_Required actions:_
1. Build codemod
2. Build test

adamviktora:
Closes #601

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/treeViewWarnSelectableStylingModifierRemoved/treeView-warn-selectable-styling-modifier-removed.ts
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/treeViewWarnSelectableStylingModifierRemoved/treeView-warn-selectable-styling-modifier-removed.ts
@@ -1,0 +1,32 @@
+import { Rule } from "eslint";
+import { JSXOpeningElement } from "estree-jsx";
+import { getFromPackage } from "../../helpers";
+
+// https://github.com/patternfly/patternfly-react/pull/10101
+module.exports = {
+  meta: {},
+  create: function (context: Rule.RuleContext) {
+    const { imports } = getFromPackage(context, "@patternfly/react-core");
+
+    const treeViewImport = imports.find(
+      (specifier) => specifier.imported.name === "TreeView"
+    );
+
+    return !treeViewImport
+      ? {}
+      : {
+          JSXOpeningElement(node: JSXOpeningElement) {
+            if (
+              node.name.type === "JSXIdentifier" &&
+              treeViewImport.local.name === node.name.name
+            ) {
+              context.report({
+                node,
+                message:
+                  "Selectable styling attribute (`pf-m-selectable`) has been removed in TreeView component.",
+              });
+            }
+          },
+        };
+  },
+};
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/treeViewWarnSelectableStylingModifierRemoved/treeView-warn-selectable-styling-modifier-removed.ts
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/treeViewWarnSelectableStylingModifierRemoved/treeView-warn-selectable-styling-modifier-removed.ts
@@ -23,7 +23,7 @@ module.exports = {
               context.report({
                 node,
                 message:
-                  "Selectable styling attribute (`pf-m-selectable`) has been removed in TreeView component.",
+                  "Selectable styling attribute (`pf-m-selectable`) has been removed on the list items of the TreeView component. You should update selectors, tests and snapshot tests which are relying on the `pf-m-selectable` class.",
               });
             }
           },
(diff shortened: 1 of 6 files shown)


packages/eslint-plugin-pf-codemods/src/rules/v6/treeViewWarnSelectableStylingModifierRemoved/treeView-warn-selectable-styling-modifier-removed.ts
```diff
@@ -0,0 +1,32 @@
...
+// https://github.com/patternfly/patternfly-react/pull/10101
+module.exports = {
+  meta: {},
+  create: function (context: Rule.RuleContext) {
+    const { imports } = getFromPackage(context, "@patternfly/react-core");
+
+    const treeViewImport = imports.find(
+      (specifier) => specifier.imported.name === "TreeView"
+    );
+
```
wise-king-sullyman, line 14: Could we add some logic so that it only throws this warning in the event that `pf-m-selectable` _would've_ been applied? 

For easy reference that logic was `children && (isSelectable || hasCheckbox)`

packages/eslint-plugin-pf-codemods/src/rules/v6/treeViewWarnSelectableStylingModifierRemoved/treeView-warn-selectable-styling-modifier-removed.ts
```diff
@@ -0,0 +1,32 @@
...
+      : {
+          JSXOpeningElement(node: JSXOpeningElement) {
+            if (
+              node.name.type === "JSXIdentifier" &&
+              treeViewImport.local.name === node.name.name
+            ) {
+              context.report({
+                node,
+                message:
+                  "Selectable styling attribute (`pf-m-selectable`) has been removed in TreeView component.",
```
wise-king-sullyman, line 26: Also could we just add a bit to the message here telling consumers what they might have to do because of this change/how it might impact them?

wise-king-sullyman (changes requested):
Looks solid, two small comments

adamviktora:
I started addressing your comments and then I realized I made a mistake in the component name. The breaking change was on a `TreeViewListItem`, not `TreeView`.

TreeViewListItem is not directly exported from `@patternfly/react-core`, we only use this component internally. Although someone may import it with `import { TreeViewListItem } from '@patternfly/react-core/dist/esm/components/TreeView/TreeViewListItem';`

Do we still need a codemod just in case someone can import it?

wise-king-sullyman:
It's still pretty much on TreeView, just indirectly.

I think a codemod would still be good though since the list item class changes will still potentially impact TreeView users. I would just worry about it in that case though and not worry about someone importing it directly like that.

adamviktora:
Oh, I see, you are right. I was way too stuck in the "fixer" mindset from all the other rules 😃 where we only do changes if the exact component is imported.

packages/eslint-plugin-pf-codemods/src/rules/v6/treeViewWarnSelectableStylingModifierRemoved/treeView-warn-selectable-styling-modifier-removed.ts
adamviktora: Do you know what the message should say? Even after checking your React PR https://github.com/patternfly/patternfly-react/pull/10101 and the original issue I have no clue

packages/eslint-plugin-pf-codemods/src/rules/v6/treeViewWarnSelectableStylingModifierRemoved/treeView-warn-selectable-styling-modifier-removed.ts
adamviktora: After doing some investigation, we are not able to check it and we will have to throw a warning every time.

The reason is simple: the `children` and `hasCheckbox` props depend on the individual TreeViewDataItem objects from `data: TreeViewDataItem[]` prop of the TreeViewProps.

And since the TreeView may be used like this:
```
import {options} from "somewhere";
<TreeView data={options} ...
```
we are not able to trace it and check the condition.

packages/eslint-plugin-pf-codemods/src/rules/v6/treeViewWarnSelectableStylingModifierRemoved/treeView-warn-selectable-styling-modifier-removed.ts
wise-king-sullyman: I would just let them know that they will need to update any selectors in their code or tests that depends on that classname being there, along with any snapshot tests that include the class.

thatblindgeye (approved):

wise-king-sullyman (approved):
🚀

PR #621, opened 2024-03-26
feat(Chip) replace with Label

### LINKED ISSUE #564: Chip/ChipGroup - replaced deprecated component with Label and LabelGroup
`Chip` and `ChipGroup` have been deprecated and it should be possible to use a codemod to update to using `Label` and `LabelGroup`.

This probably expects the codemod that updates the Chip/ChipGroup path to the new deprecated path to be run first.

Dominik-Petrik:
close #564

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/chipReplaceWithLabel/chip-replace-with-label.ts
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/chipReplaceWithLabel/chip-replace-with-label.ts
@@ -1,4 +1,9 @@
-import { getAttribute, getFromPackage } from '../../helpers';
+import {
+  getAllJSXElements,
+  getAttribute,
+  getFromPackage,
+  getImportDeclarations,
+} from '../../helpers';
 import {
   ImportDeclaration,
   ImportSpecifier,
@@ -17,41 +22,6 @@ type JSXAttributeValue =
   | JSXExpressionContainer
   | JSXElement
   | JSXFragment;
-
-const getImportSpecifier = (node: ImportDeclaration, specifierName: string) => {
-  return node.specifiers.find(
-    (specifier) =>
-      specifier.type === 'ImportSpecifier' &&
-      specifier.imported.name === specifierName
-  ) as ImportSpecifier | undefined;
-};
-
-const renameImportSpecifier = (
-  importToRename: ImportSpecifier | undefined,
-  node: ImportDeclaration,
-  fixer: Rule.RuleFixer,
-  newName: string
-) => {
-  if (!importToRename) {
-    return [];
-  }
-
-  const oldName = importToRename.imported.name;
-  const importSpecifier = getImportSpecifier(
-    node,
-    importToRename.imported.name
-  ) as ImportSpecifier;
-
-  const importedName = importSpecifier.imported.name;
-  const name = importedName.replace(oldName, newName);
-  const localName = importSpecifier.local.name;
-  const aliasText = importedName !== localName ? ` as ${localName}` : '';
-  const rename = `${name}${aliasText}`;
-  return [
-    fixer.replaceText(importSpecifier, rename),
-    fixer.replaceText(node.source, "'@patternfly/react-core'"),
-  ];
-};
 
 const parseBadgeAttributeValue = (
   badgeAttributeValue: JSXAttributeValue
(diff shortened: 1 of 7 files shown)


packages/eslint-plugin-pf-codemods/src/rules/v6/chipReplaceWithLabel/chip-replace-with-label.ts
```diff
@@ -0,0 +1,202 @@
...
+const moveBadgeAttributeToBody = (
+  node: JSXElement,
+  fixer: Rule.RuleFixer,
+  context: Rule.RuleContext
+) => {
+  const badgeAttribute = getJSXAttribute(node, 'badge');
+
+  return badgeAttribute
+    ? [
+        fixer.insertTextBefore(
```
adamviktora, line 86: Can we add a space before the text? So the result of the fix is `This is a chip {identifier}` rather than `This is a chip{identifier}`

packages/eslint-plugin-pf-codemods/src/rules/v6/chipReplaceWithLabel/chip-replace-with-label.ts
```diff
@@ -0,0 +1,202 @@
...
+                        ' variant="outline"'
+                      ),
+                      ...renameOnClickAttribute(node, fixer),
+                      ...(node.closingElement
+                        ? [
+                            fixer.replaceText(
+                              node.closingElement.name,
+                              'Label'
+                            ),
+                            ...moveBadgeAttributeToBody(node, fixer, context),
```
adamviktora, line 170: Is it possible that someone would use a Chip with Badge but without children (self-closing)? I am just wondering, whether this case should be covered:
```
<Chip onClick={onClick} badge={<Badge isRead={true}>7</Badge>} />
```
Screenshot 2024-03-26 at 16 09 06

I guess probably not, or ... it would be too much work to handle it and it is not very likely.

packages/eslint-plugin-pf-codemods/src/rules/v6/chipReplaceWithLabel/chipReplaceWithLabelInput.tsx
```diff
@@ -0,0 +1,3 @@
+import { Chip } from '@patternfly/react-core/deprecated';
+
+export const ChipReplaceWithLabelInput = () => <Chip onClick=(handleClick) badge={badge}></Chip>;
```
adamviktora, line 3: I think this was meant to be `onClick={handleClick}`.
Also can we add some text as children of the Chip? (to better demonstrate the badge transformation)

adamviktora (changes requested):
Looking great! 🔥 I like the idea of separating the fixer helper functions.

Couple of comments bellow, but mostly minor changes. The only blocking one is related to the Label / LabelGroup import - what if it is already imported, the user could end up with duplicate import specifier.

packages/eslint-plugin-pf-codemods/src/rules/v6/chipReplaceWithLabel/chip-replace-with-label.ts
```diff
@@ -0,0 +1,202 @@
...
+
+    const chipImport = imports.find(
+      (specifier) => specifier.imported.name === 'Chip'
+    );
+
+    const chipGroupImport = imports.find(
+      (specifier) => specifier.imported.name === 'ChipGroup'
+    );
+
+    return !chipImport && !chipGroupImport
```
wise-king-sullyman, line 126: Nit: I would reverse this conditional so that these expressions aren't notted and have the non-empty object returned first

packages/eslint-plugin-pf-codemods/src/rules/v6/chipReplaceWithLabel/chip-replace-with-label.ts
```diff
@@ -0,0 +1,202 @@
...
+                    fixer,
+                    'LabelGroup'
+                  ),
+                ];
+              },
+            });
+          },
+
+          JSXElement(node: JSXElement) {
+            if (node.openingElement.name.type === 'JSXIdentifier') {
```
wise-king-sullyman, line 148: Nit: I'd refactor this to a guard clause since all the rest of the logic is encapsulated in it.

packages/eslint-plugin-pf-codemods/src/rules/v6/chipReplaceWithLabel/chip-replace-with-label.ts
```diff
@@ -0,0 +1,202 @@
...
+              if (
+                chipImport &&
+                chipImport.local.name === node.openingElement.name.name
+              ) {
+                context.report({
+                  node,
+                  message: `Chip has been deprecated. Running the fix flag will replace Chip and ChipGroup components with Label and LabelGroup components respectively.`,
```
wise-king-sullyman, line 155: Nit: I'd also refactor to only have this `context.report` call once since the `fix` function is the only thing changing

wise-king-sullyman (commented):
Awesome work on this! A few small refactor nits and Adam's existing comments and I think this will be good!

packages/eslint-plugin-pf-codemods/src/rules/v6/chipReplaceWithLabel/chip-replace-with-label.ts
```diff
@@ -0,0 +1,202 @@
...
+  const importSpecifier = getImportSpecifier(
+    node,
+    importToRename.imported.name
+  ) as ImportSpecifier;
+
+  const importedName = importSpecifier.imported.name;
+  const name = importedName.replace(oldName, newName);
+  const localName = importSpecifier.local.name;
+  const aliasText = importedName !== localName ? ` as ${localName}` : '';
+  const rename = `${name}${aliasText}`;
```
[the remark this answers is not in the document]
Dominik-Petrik: Rename should be used instead of `newName` in the fixer, thanks for the catch.

packages/eslint-plugin-pf-codemods/src/rules/v6/chipReplaceWithLabel/chip-replace-with-label.ts
```diff
@@ -0,0 +1,202 @@
...
+    return !chipImport && !chipGroupImport
+      ? {}
+      : {
+          ImportDeclaration(node: ImportDeclaration) {
+            context.report({
+              node,
+              message: `Chip has been deprecated. Running the fix flag will replace Chip and ChipGroup components with Label and LabelGroup components respectively.`,
+              fix(fixer) {
+                return [
+                  ...renameImportSpecifier(chipImport, node, fixer, 'Label'),
```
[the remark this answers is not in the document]
Dominik-Petrik: Nice catch!

packages/eslint-plugin-pf-codemods/src/rules/v6/chipReplaceWithLabel/chip-replace-with-label.ts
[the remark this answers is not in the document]
adamviktora: I think you can keep the `newName` there as it can make things easier afterwards when renaming the JSXElement, so you don't have to check if an alias was used there. But it is up to you. Also the duplicate Label import might play a role, so it might end up more complicated.

adamviktora (approved):

adamviktora:
Did some changes to handle the situation with `Label` import already included. Also there was a bug when there were more components imported from the deprecated package, they were all converted to the react-core package. 

Also I discovered that when there were more elements nested, like this:
```
<ChipGroup>
  <Chip>Some chip</Chip>
</ChipGroup>
```
it would only modify the **parent element** on the first rule run (it eventually runs the rule 10 times, fixing the nested elements, but it is problematic with tests, where it only runs once) - due to some ranges clashing probably.

Because of that and also to be able to properly remove the unused Chip and ChipGroup imports (otherwise after removing, we are not able to get the previous local name - alias of the import), it was necessary to apply all the fixes in one `ImportDeclaration` matcher.

The only disadvantage is the consumers won't get errors for every line where the actual JSXElement is used, but only for the import declaration.

wise-king-sullyman (approved):
🚀

PR #625, opened 2024-04-04
feat(FormFiledGroupHeaderTitleTextObject): handle export specifier

### LINKED ISSUE #623: FormFiledGroupHeaderTitleTextObject - add checking of ExportSpecifier to codemod
Follow up to codemod https://github.com/patternfly/pf-codemods/pull/610/files

We should add onto the existing codemod to update the name of an export of `FormFiledGroupHeaderTitleTextObject`.

_Required actions:_
1. Build codemod
2. Build test

adamviktora:
Closes #623

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/formFiledGroupHeaderTitleTextObjectRenamed/formFiledGroupHeaderTitleTextObject-renamed.test.ts
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/formFiledGroupHeaderTitleTextObjectRenamed/formFiledGroupHeaderTitleTextObject-renamed.test.ts
@@ -9,6 +9,9 @@ ruleTester.run("formFiledGroupHeaderTitleTextObject-renamed", rule, {
     {
       code: `const titleTextObject: FormFiledGroupHeaderTitleTextObject = {text: "Some title", id: "form-field-group-header-title-text"};`,
     },
+    {
+      code: `export { FormFieldGroupHeaderTitleTextObject as TitleTextObject }`,
+    },
   ],
   invalid: [
     {
@@ -47,6 +50,16 @@ ruleTester.run("formFiledGroupHeaderTitleTextObject-renamed", rule, {
           message: errorMessage,
           type: "ImportSpecifier",
         },
+        {
+          message: errorMessage,
+          type: "ExportSpecifier",
+        },
+      ],
+    },
+    {
+      code: `export { FormFiledGroupHeaderTitleTextObject as TitleTextObject } from '@patternfly/react-core'`,
+      output: `export { FormFieldGroupHeaderTitleTextObject as TitleTextObject } from '@patternfly/react-core'`,
+      errors: [
         {
           message: errorMessage,
           type: "ExportSpecifier",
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin-pf-codemods/src/rules/v6/formFiledGroupHeaderTitleTextObjectRenamed/formFiledGroupHeaderTitleTextObject-renamed.ts
```diff
@@ -41,16 +41,18 @@ module.exports = {
           callContextReport(node, node.imported);
         }
       },
+      ExportSpecifier(node: ExportSpecifier) {
+        if (!importHasAlias && node.local.name === previousName) {
+          callContextReport(node, node.local);
+        }
```
thatblindgeye, line 47: Couple more things we should do for this:

- The export shouldn't be reliant on the import not being aliased, so we should remove the !importHasAlias` portion of this conditional.
- We only want to make this update if the export is from Patternfly, so we'd need to update line 9 to destructure both the imports and exports, then we want to only make an update in this ExportSpecifier block if an `interfaceExport` exists.

packages/eslint-plugin-pf-codemods/src/rules/v6/formFiledGroupHeaderTitleTextObjectRenamed/formFiledGroupHeaderTitleTextObject-renamed.test.ts
```diff
@@ -39,5 +39,19 @@ ruleTester.run("formFiledGroupHeaderTitleTextObject-renamed", rule, {
         },
       ],
     },
+    {
+      code: `import { FormFiledGroupHeaderTitleTextObject } from '@patternfly/react-core'; export { FormFiledGroupHeaderTitleTextObject as TitleTextObject }`,
+      output: `import { FormFieldGroupHeaderTitleTextObject } from '@patternfly/react-core'; export { FormFieldGroupHeaderTitleTextObject as TitleTextObject }`,
```
thatblindgeye, line 44: Related to above, for the tests we want the invalid case to be an export `from @patternfly....`, while a valid case would actually be an export like you have here (not from patternfly)

thatblindgeye (changes requested):

packages/eslint-plugin-pf-codemods/src/rules/v6/formFiledGroupHeaderTitleTextObjectRenamed/formFiledGroupHeaderTitleTextObject-renamed.ts
adamviktora: I misunderstood how the exporting can work. I thought it is necessary to first import the name to be able to reexport it, so the codemod only covers a case like this:

```
import { FormFiledGroupHeaderTitleTextObject } from "@patternfly/react-core";
export { FormFiledGroupHeaderTitleTextObject as HeaderTitleObject };
```

If I understand your review correctly, we want to cover this:
```
export { FormFiledGroupHeaderTitleTextObject as HeaderTitleObject } from "@patternfly/react-core";
```

But shouldn't we still want to have the case above covered by the codemod?

packages/eslint-plugin-pf-codemods/src/rules/v6/formFiledGroupHeaderTitleTextObjectRenamed/formFiledGroupHeaderTitleTextObject-renamed.test.ts
adamviktora: So if I understand, just this:
```
export { FormFiledGroupHeaderTitleTextObject as TitleTextObject }
```
should be valid.

But when the import is included:
```
import { FormFiledGroupHeaderTitleTextObject } from "@patternfly/react-core";
export { FormFiledGroupHeaderTitleTextObject as TitleTextObject };
```
shouldn't that be invalid? Because if the unaliased import is present, then we know the exported name is 100% the same name we import from Patternfly.

thatblindgeye (approved):

PR #648, opened 2024-05-23
feat(Toolbar): replaced spacer props with gap

### LINKED ISSUE #645: Toolbar - updates spacers to gaps
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/10418

- ToolbarGroup, ToolbarItem, and ToolbarToggleGroup: removed `spacer` prop. We could provide a fix that will replace the prop with the `gap` prop and also update the values passed in (e.g. `spacerNone` => `gapNone`, or we could just state that we recommend using our `gap`, `columnGap` or `rowGap` props instead.
- ToolbarGroup and ToolbarToggleGroup: removed the `spaceItems` prop. Will probably require just an error message stating we recommend using `marginInlineStart` and `marginInlineEnd` props on applicable components instead.

[Description of the breaking change the rules needing to be created]

_Required actions:_
1. Build codemod
2. Build test

thatblindgeye:
Closes #645

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/toolbarReplacedSpacerSpaceItems/toolbar-replaced-spacer-spaceItems.ts
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/toolbarReplacedSpacerSpaceItems/toolbar-replaced-spacer-spaceItems.ts
@@ -1,0 +1,73 @@
+import { Rule } from "eslint";
+import { JSXOpeningElement, JSXAttribute } from "estree-jsx";
+import { Property } from "estree-jsx";
+import { getFromPackage, getAttribute, getAttributeValue } from "../../helpers";
+
+// https://github.com/patternfly/patternfly-react/pull/10418
+module.exports = {
+  meta: { fixable: "code" },
+  create: function (context: Rule.RuleContext) {
+    const { imports } = getFromPackage(context, "@patternfly/react-core");
+
+    const toolbarImports = imports.filter((specifier) =>
+      ["ToolbarGroup", "ToolbarToggleGroup", "ToolbarItem"].includes(
+        specifier.imported.name
+      )
+    );
+
+    return !toolbarImports.length
+      ? {}
+      : {
+          JSXOpeningElement(node: JSXOpeningElement) {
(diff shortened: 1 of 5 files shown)


packages/eslint-plugin-pf-codemods/src/rules/v6/toolbarReplacedSpacerSpaceItems/toolbar-replaced-spacer-spaceItems.ts
```diff
@@ -0,0 +1,73 @@
+import { Rule } from "eslint";
+import { JSXOpeningElement, JSXAttribute } from "estree-jsx";
```
adamviktora, line 2: nit: unused import
```suggestion
import { JSXOpeningElement } from "estree-jsx";
```

packages/eslint-plugin-pf-codemods/src/rules/v6/toolbarReplacedSpacerSpaceItems/toolbar-replaced-spacer-spaceItems.ts
```diff
@@ -0,0 +1,73 @@
...
+                message: `${spacerProp ? spacerPropMessage : ""}${
+                  spaceItemsProp ? spaceItemsPropMessage : ""
+                }`,
+                fix(fixer) {
+                  const fixes = [];
+
+                  if (spacerProp) {
+                    spacerVal &&
+                      spacerVal.forEach((val: Property) => {
+                        const newValue =
```
adamviktora, line 51: `newValue` can be `false` here, which could be a problem in a use case like this:

```
const NO_SPACER = "spacerNone";
<ToolbarItem spacer={{ default: NO_SPACER }} />
```
which results to:
```
const NO_SPACER = "spacerNone";
<ToolbarItem spacer={{ default: "false" }} />
```

I don't think we need to attempt to change the variable value to `gapNone`, because this is a very rare use case and probably nobody uses it like this. But I think the variable name should stay there so we don't mess up with consumer's code just in case.

adamviktora (approved):
🎉

PR #649, opened 2024-05-23
feat(Banner): added color and status props

### LINKED ISSUE #647: Banner - replaced variant prop with color
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/9891

We should update the variant prop to color, unless the variant was default in which case we should remove the prop. The error message should also mention something along the lines of "if you used the variant prop for color, use color prop, otherwise use status prop if you were using variant to convey status" or something.

_Required actions:_
1. Build codemod
2. Build test

thatblindgeye:
Closes #647

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/bannerReplaceVariantProp/banner-replace-variantProp.ts
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/bannerReplaceVariantProp/banner-replace-variantProp.ts
@@ -8,17 +8,17 @@ module.exports = {
   create: function (context: Rule.RuleContext) {
     const { imports } = getFromPackage(context, "@patternfly/react-core");
 
-    const componentImports = imports.find(
+    const bannerImport = imports.find(
       (specifier) => specifier.imported.name === "Banner"
     );
 
-    return !componentImports
+    return !bannerImport
       ? {}
       : {
           JSXOpeningElement(node: JSXOpeningElement) {
             if (
               node.name.type === "JSXIdentifier" &&
-              componentImports.local.name === node.name.name
+              bannerImport.local.name === node.name.name
             ) {
               const attribute = getAttribute(node, "variant");
               if (!attribute) {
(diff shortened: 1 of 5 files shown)


packages/eslint-plugin-pf-codemods/src/rules/v6/bannerReplaceVariantProp/banner-replace-variantProp.ts
```diff
@@ -0,0 +1,51 @@
...
+import { JSXOpeningElement } from "estree-jsx";
+import { getFromPackage, getAttribute, getAttributeValue } from "../../helpers";
+
+// https://github.com/patternfly/patternfly-react/pull/9891
+module.exports = {
+  meta: { fixable: "code" },
+  create: function (context: Rule.RuleContext) {
+    const { imports } = getFromPackage(context, "@patternfly/react-core");
+
+    const componentImports = imports.find(
```
adamviktora, line 11: Could be named `bannerImport` instead.

packages/eslint-plugin-pf-codemods/src/rules/v6/bannerReplaceVariantProp/banner-replace-variantProp.ts
```diff
@@ -0,0 +1,51 @@
...
+              const isValueDefault = attributeValue === "default";
+              const fixMessage = isValueDefault
+                ? "remove the variant property"
+                : "replace the variant property with the color property";
+
+              context.report({
+                node,
+                message: `The variant property has been removed from Banner. We recommend using our new color or status properties, depending on the original intent of the variant property. Running the fix for this rule will ${fixMessage}, but additional manual updates may need to be made.`,
+                fix(fixer) {
+                  if (attributeValue === "default") {
```
adamviktora, line 40: `isValueDefault` was probably meant to be used here

adamviktora (approved):
LGTM 🎉 
(nit comments bellow, but just refactoring, can be merged like this)

PR #653, opened 2024-05-29
fix(Chip): Remove isReadOnly prop during label conversion

### LINKED ISSUE #634: Chip - isReadOnly prop does not exist on Label
It seems like when we convert a chip to a label we aren't accounting for the `isReadOnly` prop existing on the Chip but not on the Label.

wise-king-sullyman:
Closes #634


thatblindgeye (approved):

Dominik-Petrik (approved):
🚀

PR #678, opened 2024-07-01
fix(noUnusedImports): Add unused import removal for v6 rules

### LINKED ISSUE #656: The cleanup rule which removes unused imports isn't running for the v6 codemods.
I tested our v5 version of this rule in a consumers codebase and it seemed to work fine, we probably just need to do something so that it runs with the v6 rules.

wise-king-sullyman:
Closes #656


thatblindgeye (commented):
This looks good. Can we just add a test for multiple removals + 1 valid import, e.g. `import {Foo, Bar, Baz, Thing}.... <Bar />`

thatblindgeye (approved):

PR #689, opened 2024-07-09
feat(Button): updated icon usage

### LINKED ISSUE #680: Button - icons should now be passed to icon prop instead of as children
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/10663

API of Button is being tweaked, icons should always be passed to `icon` prop. Codemod should move icons when the plain variant is applied, as that is the safest assumption that only an icon is being passed in.

_Required actions:_
1. Build codemod
2. Build test

thatblindgeye:
Closes #680

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/buttonMoveIconsIconProp/button-moveIcons-icon-prop.test.ts
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/buttonMoveIconsIconProp/button-moveIcons-icon-prop.test.ts
@@ -1,0 +1,85 @@
+const ruleTester = require("../../ruletester");
+import * as rule from "./button-moveIcons-icon-prop";
+
+ruleTester.run("button-moveIcons-icon-prop", rule, {
+  valid: [
+    {
+      code: `<Button variant="plain"><span>Some icon</span></Button>`,
+    },
+    {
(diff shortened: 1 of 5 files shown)


packages/eslint-plugin-pf-codemods/src/rules/v6/buttonMoveIconsIconProp/button-moveIcons-icon-prop.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    {
+      code: `import { Button } from '@patternfly/react-core'; const icon = <span>Some icon</span>; <Button variant="plain">{icon}</Button>`,
+      output: `import { Button } from '@patternfly/react-core'; const icon = <span>Some icon</span>; <Button icon={icon} variant="plain"></Button>`,
+      errors: [
+        {
+          message: `Icons must now be passed to the \`icon\` prop of Button instead of as children. If you are passing anything other than an icon as children, ignore this rule when running fixes.`,
+          type: "JSXElement",
+        },
+      ],
+    },
```
wise-king-sullyman, line 23: Assuming that all children will always be removed by the codemod (which I _think_ is the case when it meets the expected criteria for the codemod to run) I think it would be nice if we converted it to a self closing tag, but I wouldn't block over that.

wise-king-sullyman (approved):

PR #691, opened 2024-07-09
feat(Banner,Label): renamed cyan/gold color values

### LINKED ISSUE #682: Banner - updated colors to new names
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/10661

We will need to autofix the old colors to new colors

_Required actions:_
1. Build codemod
2. Build test

thatblindgeye:
Closes #682

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/colorPropsReplacedColors/colorProps-replaced-colors.test.ts
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/colorPropsReplacedColors/colorProps-replaced-colors.test.ts
@@ -1,0 +1,112 @@
+const ruleTester = require("../../ruletester");
+import * as rule from "./colorProps-replaced-colors";
+import {
+  ValidTests,
+  InvalidTests,
+  createInvalidTest,
+  createValidTest,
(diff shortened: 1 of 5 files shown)


packages/eslint-plugin-pf-codemods/src/rules/v6/colorPropsReplacedColors/colorProps-replaced-colors.test.ts
```diff
@@ -0,0 +1,112 @@
...
+          message: `The \`color\` prop on ${component} has been updated to replace "cyan" with "teal".`,
+          type,
+        },
+      ]
+    )
+  );
+  invalidTests.push(
+    createInvalidTest(
+      `import { ${component} } from '@patternfly/react-core'; const color = 'gold'; <${component} color={color} />`,
+      `import { ${component} } from '@patternfly/react-core'; const color = 'gold'; <${component} color="yellow" />`,
```
wise-king-sullyman, line 50: I think we would ideally update the assignment in situations like this rather than just the use, not a blocker though.

wise-king-sullyman (approved):

PR #702, opened 2024-07-18
feat(Dropdown deprecated): added rule to replace with MenuToggle

### LINKED ISSUE #683: KebabToggle - Address KebabToggle removal
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/10345

The KebabToggle was removed, ideally we should provide a fixer that swaps in a MenuToggle with the appropriate props and an EllipsisVIcon, but at a minimum we should throw an error to tell consumers they need to replace it and give them a push in the right direction.

_Required actions:_
1. Build codemod
2. Build test

thatblindgeye:
Closes #683 

Also includes some cleanup/helpers added in.


packages/eslint-plugin-pf-codemods/src/rules/v6/kebabToggleReplaceWithMenuToggle/kebabToggle-replace-with-menuToggle.test.ts
```diff
@@ -0,0 +1,188 @@
...
+      errors: [
+        {
+          message,
+          type: "ImportDeclaration",
+        },
+      ],
+    },
+    {
+      code: `import { KebabToggle } from "@patternfly/react-core/dist/js/deprecated/components/Dropdown/index.js";`,
+      output: `import { MenuToggle } from "@patternfly/react-core";\nimport EllipsisVIcon from "@patternfly/react-icons/dist/esm/icons/ellipsis-v-icon";`,
```
wise-king-sullyman, line 155: 🤔 this may or may not really be an issue, but I think we should import the icon from /js if the consumer is importing stuff from /js already.

We may also need to reevaluate if importing from /esm by default is a safe assumption.... but I wouldn't block over that.

packages/eslint-plugin-pf-codemods/src/rules/v6/kebabToggleReplaceWithMenuToggle/kebabToggle-replace-with-menuToggle.test.ts
```diff
@@ -0,0 +1,188 @@
...
+      errors: [
+        {
+          message,
+          type: "ImportDeclaration",
+        },
+      ],
+    },
+    {
+      code: `import { KebabToggle } from "@patternfly/react-core/dist/dynamic/deprecated/components/Dropdown/index.js";`,
```
wise-king-sullyman, line 164: Probably the same for /dynamic

wise-king-sullyman (approved):

packages/eslint-plugin-pf-codemods/src/rules/v6/kebabToggleReplaceWithMenuToggle/kebabToggle-replace-with-menuToggle.test.ts
wise-king-sullyman: Made #705 as a followup

PR #710, opened 2024-07-23
feat(LogSnippet): rename leftBorderVariant prop, remove enum

### LINKED ISSUE #668: [component-groups | LogSnippet] - Rename leftBorderVariant
Follow up to breaking change [PR](https://github.com/patternfly/react-component-groups/pull/145)

We've renamed LogSnippet prop because of design changes and switching from border variant to alert variant 

rename
`leftBorderVariant` to `variant`

The `LogSnippetBorderVariant` has been removed and replaced with PF `AlertVariant`.  Perhaps the current values could be mapped to the new ones since the types almost match

Example:
```
<LogSnippet message='Failure - check logs for details' logSnippet={code} leftBorderVariant={LogSnippetBorderVariant.success} />
```
becomes
```
<LogSnippet message='Failure - check logs for details' logSnippet={code} variant={AlertVariant.success} />
```
or
```
<LogSnippet message='Failure - check logs for details' logSnippet={code} variant="sucess" />
```

Required actions:
- Build codemod
- Build test
- Update readme with description & example

adamviktora:
Closes #668


packages/eslint-plugin-pf-codemods/src/rules/helpers/getFromPackage.ts
```diff
@@ -103,6 +104,10 @@ export function getDefaultImportsFromPackage(
   ) as ImportDeclaration[];
 
   return importDeclarationsFromPackage
-    .filter((imp) => imp.specifiers[0]?.type === "ImportDefaultSpecifier")
+    .filter(
+      (imp) =>
+        imp.source.value?.toString().includes(componentName) &&
```
thatblindgeye, line 109: Would we always want to  run this `includes` check, or only if componentName is passed a non-empty value?

thatblindgeye (commented):
Just one question below, otherwise lgtm

packages/eslint-plugin-pf-codemods/src/rules/helpers/getFromPackage.ts
adamviktora: Yeah, that should run always. The idea was that if no value is passed, the includes check will return true. But it is an unnecessary check, let me rewrite that

packages/eslint-plugin-pf-codemods/src/rules/v6/componentGroups-logSnippetRenameLeftBorderVariant/componentGroups-logSnippet-rename-leftBorderVariant.ts
```diff
@@ -0,0 +1,93 @@
...
+              logSnippetBorderVariantImports
+                .map((imp) => imp.local.name)
+                .includes(node.object.name)
+            ) {
+              const replaceToken =
+                node.property.type === "Identifier"
+                  ? node.property.name
+                  : node.property.type === "Literal"
+                  ? (node.property.value as string)
+                  : null;
```
wise-king-sullyman, line 68: Super small nit, but IMO a quick switch here would be a lot cleaner than a ternary chain

PR #712, opened 2024-07-23
feat(MultiContentCard): Remove leftBorderVariant and withHeaderBorder

### LINKED ISSUE #669: [component-groups | MultiContentCard] - Remove props
Follow up to breaking change [PR](https://github.com/patternfly/react-component-groups/pull/145)

We've removed MultiContentCard props because of design changes in v6

remove `leftBorderVariant` and `withHeaderBorder` 

We don't display borders anymore.

Example:
```
<MultiContentCard cards={cards} leftBorderVariant="danger" withHeaderBorder />
```
becomes
```
<MultiContentCard cards={cards} />
```

Required actions:
- Build codemod
- Build test
- Update readme with description & example

wise-king-sullyman:
Closes #669

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/rules/helpers/renamePropsOnNode.ts
+++ b/packages/eslint-plugin-pf-codemods/src/rules/helpers/renamePropsOnNode.ts
@@ -1,33 +1,66 @@
-import { Rule } from 'eslint';
+import { Rule } from "eslint";
 import {
   JSXOpeningElement,
   ImportSpecifier,
   JSXAttribute,
   JSXIdentifier,
-} from 'estree-jsx';
-import { renameSinglePropOnNode, Renames } from './renameSinglePropOnNode';
+  ImportDefaultSpecifier,
+  ImportDeclaration,
+} from "estree-jsx";
+import { renameSinglePropOnNode, Renames } from "./renameSinglePropOnNode";
+
+interface ImportDefaultSpecifierWithParent extends ImportDefaultSpecifier {
+  parent?: ImportDeclaration;
+}
+
+function getComponentName(
+  specifier: ImportSpecifier | ImportDefaultSpecifierWithParent,
+  renames: Renames
+) {
+  if (specifier.type === "ImportSpecifier") {
+    return specifier.imported.name;
+  }
+
+  const localName = specifier.local.name;
+  if (renames[localName]) {
+    return localName;
+  }
+
+  const renamedComponents = Object.keys(renames);
+  const importString = specifier?.parent?.source.value?.toString();
+
+  for (const component of renamedComponents) {
+    if (importString?.includes(component)) {
+      return component;
+    }
+  }
+
+  return "";
+}
 
 export function renamePropsOnNode(
   context: Rule.RuleContext,
-  imports: ImportSpecifier[],
+  imports: (ImportSpecifier | ImportDefaultSpecifier)[],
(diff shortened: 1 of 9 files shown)


packages/eslint-plugin-pf-codemods/src/rules/helpers/pfPackageMatches.ts
```diff
@@ -16,9 +16,11 @@ export function pfPackageMatches(
...
       "(/dist/(esm|js|dynamic))?" +
       (parts[2] ? "/" + parts[2] : "") +
-      `(/(components|helpers${
-        parts[1] === "react-icons" ? "|icons" : ""
-      })/.*)?$`
+      (parts[1] === "react-component-groups"
+        ? `(/.*)?$`
+        : `(/(components|helpers${
+            parts[1] === "react-icons" ? "|icons" : ""
+          })/.*)?$`)
```
wise-king-sullyman, line 23: Copy of what was done in #710 to minimize conflicts, though I really feel like this helper should be cleaned up as this dynamic regex is getting _really_ gnarly.

packages/eslint-plugin-pf-codemods/src/rules/v6/componentGroupsMultiContentCardRemoveProps/component-groups-multi-content-card-remove-props.ts
```diff
@@ -0,0 +1,12 @@
...
+
+// https://github.com/patternfly/react-component-groups/pull/145
+
+const renames = {
+  MultiContentCard: { leftBorderVariant: "", withHeaderBorder: "" },
+};
+
+module.exports = {
+  meta: { fixable: "code" },
+  create: renameProps(renames, "@patternfly/react-component-groups"),
```
adamviktora, line 11: I think using the `renameProps` helper won't work with default imports. These components in react-component-groups have also default exports so they can be imported like this:

`import MyCard from '@patternfly/react-component-groups/dist/dynamic/MultiContentCard';`

I ran to this issue yesterday and I first wanted to rewrite the `getFromPackage` helper, which is used internally by `renameProps` to also handle ImportDefaultSpecifiers, but it would be too painful, so I decided to do a custom codemod instead (https://github.com/patternfly/pf-codemods/pull/711/files#diff-24111f58cfc0f6b2a81d0fa8599c5b82889b4f88666e8c4e8897694235a6ca59)

packages/eslint-plugin-pf-codemods/src/rules/helpers/renamePropsOnNode.ts
```diff
@@ -1,33 +1,66 @@
...
   JSXAttribute,
   JSXIdentifier,
-} from 'estree-jsx';
-import { renameSinglePropOnNode, Renames } from './renameSinglePropOnNode';
+  ImportDefaultSpecifier,
+  ImportDeclaration,
+} from "estree-jsx";
+import { renameSinglePropOnNode, Renames } from "./renameSinglePropOnNode";
+
+interface ImportDefaultSpecifierWithParent extends ImportDefaultSpecifier {
```
thatblindgeye, line 12: Not sure if we'd ever re-use it, but may not hurt to move this into the interfaces helper file

thatblindgeye (approved):

adamviktora (approved):
Awesome work on fixing the helper!

PR #713, opened 2024-07-24
feat: No duplicate import specifiers rule

### LINKED ISSUE #709: Create a general no-duplicate-imports rule
Create a rule that will run as a last rule (similar to `no-unused-imports` rule) and it will get rid of duplicate imports (same imported and local name from the same package).

_Required actions:_
1. Build codemod
2. Build test

adamviktora:
Closes #709

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/noDuplicateImportSpecifiers/no-duplicate-import-specifiers.test.ts
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/noDuplicateImportSpecifiers/no-duplicate-import-specifiers.test.ts
@@ -1,0 +1,64 @@
+const ruleTester = require("../../ruletester");
+import * as rule from "./no-duplicate-import-specifiers";
+
+ruleTester.run("no-duplicate-import-specifiers", rule, {
+  valid: [
+    {
+      // we care only about imports from "@patternfly/react-core"
+      code: `import { Button, Button } from "somewhere"`,
+    },
+    {
+      code: `import { Button, Button as AnotherButton } from "@patternfly/react-core";
+      <>
+        <Button>Sample button</Button>
+        <AnotherButton>Another one</AnotherButton>
+      </>`,
+    },
+  ],
+  invalid: [
+    {
+      code: `import { Button, Button } from "@patternfly/react-core";
+      <Button>Sample button</Button>`,
+      output: `import { Button,  } from "@patternfly/react-core";
+      <Button>Sample button</Button>`,
+      errors: [
+        {
+          message: `Duplicate import specifier Button imported from '@patternfly/react-core'.`,
+          type: "ImportSpecifier",
+        },
+      ],
+    },
+    {
+      code: `import { Button } from "@patternfly/react-core";
+      import { Button } from "@patternfly/react-core";
+      <Button>Sample button</Button>`,
+      output: `import { Button } from "@patternfly/react-core";
+      
+      <Button>Sample button</Button>`,
+      errors: [
+        {
+          message: `Duplicate import specifier Button imported from '@patternfly/react-core'.`,
+          type: "ImportSpecifier",
+        },
+      ],
+    },
+    {
+      code: `import { Button as BTN, TextInput, Button as BTN } from "@patternfly/react-core";
+      <>
+        <BTN>Sample button</BTN>
+        <TextInput>Text</TextInput>
+      </>`,
+      output: `import { Button as BTN, TextInput,  } from "@patternfly/react-core";
+      <>
+        <BTN>Sample button</BTN>
+        <TextInput>Text</TextInput>
+      </>`,
+      errors: [
+        {
+          message: `Duplicate import specifier BTN imported from '@patternfly/react-core'.`,
+          type: "ImportSpecifier",
+        },
+      ],
+    },
+  ],
+});
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/noDuplicateImportSpecifiers/no-duplicate-import-specifiers.test.ts
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/noDuplicateImportSpecifiers/no-duplicate-import-specifiers.test.ts
@@ -14,6 +14,22 @@ ruleTester.run("no-duplicate-import-specifiers", rule, {
         <AnotherButton>Another one</AnotherButton>
       </>`,
     },
+    {
+      code: `import { Select } from '@patternfly/react-core/deprecated';
+      import { Select } from '@patternfly/react-core';`
+    },
+    {
+      code: `import { Select } from '@patternfly/react-core/deprecated';
+      import { Select } from '@patternfly/react-core/dist/dynamic/components/Select';`
+    },
+    {
+      code: `import { Select } from '@patternfly/react-core/next';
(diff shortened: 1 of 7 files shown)


packages/eslint-plugin-pf-codemods/src/rules/v6/noDuplicateImportSpecifiers/no-duplicate-import-specifiers.test.ts
```diff
@@ -0,0 +1,64 @@
...
+import * as rule from "./no-duplicate-import-specifiers";
+
+ruleTester.run("no-duplicate-import-specifiers", rule, {
+  valid: [
+    {
+      // we care only about imports from "@patternfly/react-core"
+      code: `import { Button, Button } from "somewhere"`,
+    },
+    {
+      code: `import { Button, Button as AnotherButton } from "@patternfly/react-core";
```
wise-king-sullyman, line 11: Should we preserve imports of the same thing with different local names like this? 🤔

As an example I feel like if someone has `import { Text, TextContent as TC}` it should become `import { Content }` and the local name removed, not `import { Content, Content as TC }`.

I won't block either way though since it is valid JS.

packages/eslint-plugin-pf-codemods/src/rules/v6/noDuplicateImportSpecifiers/no-duplicate-import-specifiers.ts
```diff
@@ -0,0 +1,78 @@
...
+
+        if (seen.has(key)) {
+          duplicates.push(specifier);
+        } else {
+          seen.add(key);
+        }
+      });
+
+      return duplicates;
+    };
```
wise-king-sullyman, line 45: This is very cool, but also feels a bit complex for what we're really doing here.

Could we do something like:
```
    const findDuplicates = (specifiers: ImportSpecifier[]) => {
      const localNames = specifiers.map((spec) => spec.local.name);

      return specifiers.filter(
        (specifier, index) => localNames.indexOf(specifier.local.name) !== index
      );
    };

Comment (adamviktora) [reply packages/eslint-plugin-pf-codemods/src/rules/v6/noDuplicateImportSpecifiers/no-duplicate-import-specifiers.ts:45]: Oh, that looks much nicer. 

I was for some reason stuck on a mindset, that both `imported` and `local` names must match. Because if consumer would have something like: `import { Button as ABC, Text as ABC }` we would remove an import of a component we don't want to remove - but such code is invalid in the first place, so we can work with local names only.

File: packages/eslint-plugin-pf-codemods/src/rules/v6/noDuplicateImportSpecifiers/no-duplicate-import-specifiers.test.ts (hunk 1 above)
Comment (wise-king-sullyman) [reply packages/eslint-plugin-pf-codemods/src/rules/v6/noDuplicateImportSpecifiers/no-duplicate-import-specifiers.test.ts:11]: Ah good point!

File: packages/eslint-plugin-pf-codemods/src/rules/v6/textReplaceWithContent/textReplaceWithContentOutput.tsx [hunk 3]
```diff
@@ -1,7 +1,4 @@
 import {
-  Content,
-  Content,
-  Content,
```
wise-king-sullyman, line 4: Should this unrelated rules output be updated? Or are we keeping the *Output.tsx files as the output of running _just_ the associated rule on the *Input.tsx file?

adamviktora: That's a good question. I maybe lean a bit more towards having it as the output of running all rules on the file. But I don't know how we approached this in other rules. What is your opinion on this? Also tagging @thatblindgeye

wise-king-sullyman: I lean more towards having it the output of just the associated rule since these files get pulled into the readme as depictions of what each rule will do, and that's at least what I've been doing and what I _thought_ yall were doing.

Something that's been talked about in the past is the idea of compiling all of input/output files together into master input/output files to ensure that rules play nice together, would you feel better about the "just the associated rule" approach if we rolled out something like that?

thatblindgeye: I had been going with having each individual Output file only apply fixes for that specific rule. A "master" file of all in/outputs may not be bad, assuming it's more just for our own eyes (and not something we publish in any README or anything). That's basically what we used to do for v5 was the single file where we added everything to ensure all the rules that ran didn't conflict with one another (too badly).

wise-king-sullyman: Yeah in my head at least it would just be for us to look at, and _maybe_ to build a new CI check around 🤔

thatblindgeye: A CI check would be less tedious for us to have to constantly check that master file and *maybe* notice any issues

adamviktora: I reverted the file so it is not updated by the no-duplicate-imports rule.

We may have the master file + CI check, I don't find it super necessary, but if there will be time for that, why not, created [this issue](https://github.com/patternfly/pf-codemods/issues/721) for it.

wise-king-sullyman (approved):

packages/eslint-plugin-pf-codemods/src/rules/v6/noDuplicateImportSpecifiers/no-duplicate-import-specifiers.test.ts
```diff
@@ -0,0 +1,64 @@
...
+      errors: [
+        {
+          message: `Duplicate import specifier Button imported from '@patternfly/react-core'.`,
+          type: "ImportSpecifier",
+        },
+      ],
+    },
+    {
+      code: `import { Button } from "@patternfly/react-core";
+      import { Button } from "@patternfly/react-core";
```
thatblindgeye, line 33: Could we also add an invalid test where Component is imported from react-core, but also imported from react-core/dist/...

packages/eslint-plugin-pf-codemods/src/rules/v6/noDuplicateImportSpecifiers/no-duplicate-import-specifiers.test.ts
```diff
@@ -0,0 +1,64 @@
+const ruleTester = require("../../ruletester");
+import * as rule from "./no-duplicate-import-specifiers";
+
+ruleTester.run("no-duplicate-import-specifiers", rule, {
+  valid: [
+    {
+      // we care only about imports from "@patternfly/react-core"
+      code: `import { Button, Button } from "somewhere"`,
```
thatblindgeye, line 8: And maybe a valid test where Component is imported from react-core, but also react-core/[deprecated|next]?

thatblindgeye (approved):
✨

PR #719, opened 2024-07-26
chore(generators): made enhancements to generator files

### LINKED ISSUE #720: Add enhancements to generators for dynamic url and cleaning up generic rule code
The generators could use some enhancing, namely:

- Since we are now creating rules specific to component groups, we should dynamically create the correct PR url for individual README files and the commented out url in code files (rather than needing to manually update these URLs)
- Cleanup the rule code for the generic rule to use helpers instead of e.g. "node.attributes.find....`

thatblindgeye:
Closes #720

### DIFF
--- a/generators/src/plop-interfaces.ts
+++ b/generators/src/plop-interfaces.ts
@@ -2,6 +2,7 @@ export interface Answers {
   componentName: string;
   propName: string;
   ruleName: string;
+  referenceRepo: string;
   referencePR: string;
   message?: string;
 }
--- a/generators/src/write-rule.ts
+++ b/generators/src/write-rule.ts
@@ -22,15 +22,16 @@ export async function genericRule({
   componentName,
   propName,
   ruleName,
+  referenceRepo,
   referencePR,
   message,
 }: Answers) {
   // the formatting for content here looks weird, but that's to preserve indentation in the written file
   const content = `import { Rule } from "eslint";
 import { JSXOpeningElement } from "estree-jsx";
-import { getFromPackage } from "../../helpers";
+import { getFromPackage, checkMatchingJSXOpeningElement, getAttribute } from "../../helpers";
 
-// https://github.com/patternfly/patternfly-react/pull/${referencePR}
+// https://github.com/patternfly/${referenceRepo}/pull/${referencePR}
 module.exports = {
   meta: { fixable: "code" },
   create: function (context: Rule.RuleContext) {
@@ -44,16 +45,8 @@ module.exports = {
       ? {}
       : {
           JSXOpeningElement(node: JSXOpeningElement) {
-            if (
-              node.name.type === "JSXIdentifier" &&
-              componentImports
-                .map((imp) => imp.local.name)
-                .includes(node.name.name)
-            ) {
-              const attribute = node.attributes.find(
-                (attr) =>
-                  attr.type === "JSXAttribute" && attr.name.name === "${propName}"
-              );
+            if (checkMatchingJSXOpeningElement(node, componentImports)) {
+              const attribute = getAttribute(node, "${propName}");
               if (attribute) {
                 context.report({
                   node,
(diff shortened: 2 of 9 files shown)


packages/eslint-plugin-pf-codemods/src/rules/v6/accordionItemWarnUpdateMarkup/accordionItem-warn-update-markup.ts
```diff
@@ -17,10 +17,7 @@ module.exports = {
       ? {}
       : {
           JSXOpeningElement(node: JSXOpeningElement) {
-            if (
-              node.name.type === "JSXIdentifier" &&
-              accordionItemImport.local.name === node.name.name
-            ) {
+            if (checkMatchingJSXOpeningElement(node, accordionItemImport)) {
```
thatblindgeye, line 20: Only updated the one file for now. I could update the rest of the files as part of this or we can do that as a followup at some other point, I don't mind either way

wise-king-sullyman: I'm also good either way you want to go about it.

generators/src/plop-interfaces.ts
```diff
@@ -2,6 +2,7 @@ export interface Answers {
   componentName: string;
   propName: string;
   ruleName: string;
+  referenceRepo: string;
```
wise-king-sullyman, line 5: Should we maybe have this be optional, with it defaulting to pf-react since that is what _most_ of our mods are for?

wise-king-sullyman (commented):
Love this, just one question/thought

generators/src/plop-interfaces.ts
thatblindgeye: @adamviktora wdyt?

For context there was some discussion on Slack:

- My intent was to have the enum sort of be the source of truth for any repo names we create codemods for, with that being the only thing we would have to update (optimistically at least; the prompt creates an array for the values of this enum). It would involve less manually typing an exact repo correctly/ensuring the repo name is exactly correct for the PR url we create for the files
- Alternatively, the repo name prompt becomes a regular input that defaults to `patternfly-react`, but have the ability to input custom text if you want

packages/eslint-plugin-pf-codemods/src/rules/v6/accordionItemWarnUpdateMarkup/accordionItem-warn-update-markup.ts
adamviktora: I don't think we need to refactor our existing codemods, since they are a "one time use" code and we probably won't copy the code from them in the future. So IMO updating the `write-rule` generator is sufficient.

generators/src/write-rule.ts
```diff
@@ -44,16 +45,8 @@ module.exports = {
...
-              node.name.type === "JSXIdentifier" &&
-              componentImports
-                .map((imp) => imp.local.name)
-                .includes(node.name.name)
-            ) {
-              const attribute = node.attributes.find(
-                (attr) =>
-                  attr.type === "JSXAttribute" && attr.name.name === "${propName}"
-              );
+            if (checkMatchingJSXOpeningElement(node, componentImports)) {
```
adamviktora, line 48: Just a question - do we want to also care about default imports? It was a concern in the `react-component-groups` repo, where I created the `getImportSpecifiersLocalNames` helper for that, which was using the `getDefaultImportsFromPackage` helper inside. But it is rather an exception and most of the rules will still be on the `patternfly-react` repo where we don't have the default imports. So I am ok with keeping the generic generator like this.

adamviktora (approved):

generators/src/write-rule.ts
wise-king-sullyman: Yeah it isn't something we need in most cases, but is there any reason we _wouldn't_ want to handle default imports by default though? 🤔

wise-king-sullyman (approved):

wise-king-sullyman (approved):
🚀

Issue #708, opened 2024-07-22
Text to Content codemod - apply the fix in one run (handle nested JSXElements)


adamviktora:
Fix the `text-replace-with-content` rule in a way so the fix is applied in 1 run so the currently commented tests pass and it works correctly for > 10 nested elements

PR #726, opened 2024-07-31
fix(textReplaceWithContent): fix nested JSXElements handling

### LINKED ISSUE #724: Text -> Content mod doesn't work in several instances in user feedback
In https://github.com/patternfly/react-user-feedback/blob/main/packages/module/src/Feedback/FeedbackForm.tsx along with the FeedbackError, FeedbackModalInternal, and FeedbackSuccess files the TextContent is changed but the child Text is not:

image

adamviktora:
Closes #724 and #708


wise-king-sullyman (approved):
🥳

thatblindgeye (approved):

PR #728, opened 2024-07-31
feat(buttonMoveIconsIconProp rule): transform to self closing tag

### LINKED ISSUE #694: buttonMoveIconsIconProp rule: transform to self closing tag
Followup to https://github.com/patternfly/pf-codemods/pull/689#discussion_r1673034800

We should try to make the component self closing, since we are moving any children to the `icon` prop and no longer need the closing tag.

_Required actions:_
1. Build codemod
2. Build test

adamviktora:
Closes #694 

Also supports Icon / react-icons being used in children attribute.

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/buttonMoveIconsIconProp/button-moveIcons-icon-prop.ts
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/buttonMoveIconsIconProp/button-moveIcons-icon-prop.ts
@@ -1,6 +1,7 @@
 import { Rule } from "eslint";
-import { JSXElement } from "estree-jsx";
+import { JSXElement, JSXFragment } from "estree-jsx";
 import {
+  childrenIsEmpty,
   getFromPackage,
   getAttribute,
   getAttributeValue,
@@ -8,6 +9,7 @@ import {
   getChildrenAsAttributeValueText,
   getChildJSXElementByName,
   isReactIcon,
+  makeJSXElementSelfClosing,
 } from "../../helpers";
 
 // https://github.com/patternfly/patternfly-react/pull/10663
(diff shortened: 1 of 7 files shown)


packages/eslint-plugin-pf-codemods/src/rules/v6/buttonMoveIconsIconProp/button-moveIcons-icon-prop.ts
```diff
@@ -87,6 +87,18 @@ module.exports = {
...
+                  if (node.closingElement) {
+                    const closingSymbol = context
+                      .getSourceCode()
+                      .getLastToken(node.openingElement)!;
+
+                    fixes.push(
+                      fixer.replaceText(closingSymbol, " />"),
+                      fixer.remove(node.closingElement)
+                    );
+                  }
```
wise-king-sullyman, line 100: Think it would be worth helperizing this? Seems like something that would be useful in a lot of situations.

Though if we do helperize it we should probably add a check to only make an element self closing if it doesn't have any children.

wise-king-sullyman (approved):
Looks great! One non-blocking thought.

thatblindgeye (approved):
Other than the above lgtm

packages/eslint-plugin-pf-codemods/src/rules/v6/buttonMoveIconsIconProp/button-moveIcons-icon-prop.ts
adamviktora: I added the helper.

About checking that element has no children - this was problematic, because for example in this button rule, we remove the children in a fix and at the same time this fix for removing closing element is applied (so the node still has children at the time the condition is checked). We would have to run the self-closing fix additionaly in a cleanup rule.

So I made the helper customizable where it is not allowed to remove children by default, but it can be overwritten.

adamviktora:
We should wait with merging https://github.com/patternfly/pf-codemods/pull/735 first. There will be conflicts and the helper will need updates, because we cannot remove all children content in case of non `plain` Buttons.

thatblindgeye (approved):

Issue #718, opened 2024-07-26
Masthead -Updates for Mastheads new structure


tlabaj:
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/10809

The structure os the Masthead has changed.  The code mode should do the following in ORder

1. Renamed `MastheadBrand` to `MastheadLogo`
2. Renamed `MastheadMain` to `MastheadBrand`
3. Wrap `MastheadToggle` and `MastheadBrand` in `MastheadMain`

_Required actions:_
1. Build codemod
2. Build test
3. Update readme with description & example

PR #729, opened 2024-07-31
feat(Masthead): Update subcomponent names

wise-king-sullyman:
Towards #718


packages/eslint-plugin-pf-codemods/src/rules/v6/mastheadNameChanges/masthead-name-changes.test.ts
```diff
@@ -0,0 +1,119 @@
...
+        {
+          message: `MastheadBrand has been renamed to MastheadLogo.`,
+          type: "JSXOpeningElement",
+        },
+      ],
+    },
+    // dist imports
+    {
+      code: `import { MastheadBrand } from '@patternfly/react-core/dist/esm/components/Masthead/MastheadBrand'; <MastheadBrand />`,
+      output: `import { MastheadLogo } from '@patternfly/react-core/dist/esm/components/Masthead/MastheadBrand'; <MastheadLogo data-codemods />`,
```
adamviktora, line 90: Shouldn't the import path be changed to `.../Masthead/MastheadLogo` ? I am not sure about this, just asking.

packages/eslint-plugin-pf-codemods/src/rules/v6/mastheadNameChanges/masthead-name-changes.md
```diff
@@ -0,0 +1,18 @@
+### masthead-name-changes [(#10809)](https://github.com/patternfly/patternfly-react/pull/10809)
+
+The old `MastheadBrand` component has been renamed to `MastheadLogo`, and the old `MastheadMain` has been renamed to `MastheadBrand`.
```
adamviktora, line 3: I think the message might be confusing for some consumers (or at least the messages given in the codemod output "MastheadMain has been renamed to MastheadBrand"), I like using the word "old" here, which makes more sense.

I was confused at the first place why are we renaming these components, when they will still stay in the new Masthead.

Maybe we can add a sentence here, like "`MastheadMain will now wrap both MastheadToggle and MastheadBrand within itself.`".

I know there will probably be a second codemod for wrapping the components in MastheadMain and the message there will better describe the whole change, but still this renaming which is not really renaming might be confusing (but probably not many consumers read the messages, I guess they just run the fix, so I am ok with keeping it as is).

adamviktora (commented):
This is looking great! 

I like the new helpers and the way they are structured + their description, it is a good starting point for refactoring the other helpers too.

Some comments / questions bellow

packages/eslint-plugin-pf-codemods/src/rules/helpers/getDeclarationString.ts
```diff
@@ -0,0 +1,8 @@
+import { ImportDefaultSpecifierWithParent } from "./interfaces";
+
+/** Gets the import path string for a default import */
+export function getDeclarationString(
```
thatblindgeye, line 4: More of a nit so not a blocker, but would `getDefaultDeclarationString` or something similar be more accurate?

packages/eslint-plugin-pf-codemods/src/rules/helpers/getNodeName.ts
```diff
@@ -0,0 +1,23 @@
+import { JSXOpeningElement, JSXMemberExpression } from "estree-jsx";
+
+/** Gets the name of an opening element or member expression */
+export function getNodeName(node: JSXOpeningElement | JSXMemberExpression) {
```
thatblindgeye, line 4: Going off of Adam's comment above, another nit so not a blocker but if this helper is meant solely for these 2 scenarios, might be worth updating the name to reflect that (if we can without it become tooooo verbose).

packages/eslint-plugin-pf-codemods/src/rules/helpers/hasCodemodDataTag.ts
```diff
@@ -0,0 +1,13 @@
+import { JSXAttribute, JSXOpeningElement } from "estree-jsx";
+import { getAttributeName } from "./getAttributeName";
+
+/** Returns true if the passed opening element has a data-codemods attribute */
+export function hasCodeModDataTag(openingElement: JSXOpeningElement) {
```
thatblindgeye, line 5: Could always be a separate helper, but just wondering if we should only check if the data tag exists by name, or return the actual attribute and base conditional logic on that. Just thinking if we ever start supplying specific values to a `data-codemods` attribute, e.g. maybe we only want to skip/run a rule if an openingElement has `data-codemods="renamed"`, but not `data-codemods="deprecated"` or something.

packages/eslint-plugin-pf-codemods/src/rules/helpers/getNodeName.ts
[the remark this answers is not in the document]
wise-king-sullyman: So my idea with just naming this helper `getNodeName` was that over time we would expand it for resolving the name of nodes of various types.

I'm not against splitting the logic into discrete private functions and minimizing recursion to help improve readability at all though 🙂 

CC @thatblindgeye since you left a comment about this same kind of stuff.

packages/eslint-plugin-pf-codemods/src/rules/v6/mastheadNameChanges/mastheadNameChangesOutput.tsx
```diff
@@ -0,0 +1,7 @@
+import { MastheadLogo, MastheadBrand } from "@patternfly/react-core";
```
thatblindgeye, line 1: This output just needs to be re-updated, looks like it's still using the fixed output from when you were also updating MastheadMain to MastheadBrand

thatblindgeye (commented):
Just one comment below. I'm good with the getNodeName helper name based on your explanation above.

packages/eslint-plugin-pf-codemods/src/rules/helpers/renameComponent.ts
```diff
@@ -0,0 +1,117 @@
...
+      fixer.replaceText(importDeclaration.source, newImportDeclaration)
+    );
+  }
+
+  const shouldRenameNode =
+    isNamedImport && nodeImport.imported.name === nodeImport.local.name;
+
+  if (shouldRenameNode) {
+    fixes.push(fixer.replaceText(node.name, newName));
+    fixes.push(fixer.insertTextAfter(node.name, " data-codemods"));
```
adamviktora, line 57: I wonder if this should be optional, I know we need the `data-codemods` for the followup Masthead rule, but in other cases it may be an unnecessary addition.

But if anything, it can be done as a followup.

adamviktora (approved):
🎉

packages/eslint-plugin-pf-codemods/src/rules/v6/mastheadNameChanges/mastheadNameChangesOutput.tsx
wise-king-sullyman: Whoops

packages/eslint-plugin-pf-codemods/src/rules/helpers/renameComponent.ts
wise-king-sullyman: Yeah I went back and forth on that and IIRC talked it over with @thatblindgeye, I think we landed on always adding it to prevent issues and adding https://github.com/patternfly/pf-codemods/issues/723 so that we aren't leaving consumer codebases littered with our data tags.

I'm not super opposed to making it optional though, but yeah would want to do it as a followup.

thatblindgeye (approved):

PR #735, opened 2024-08-09
feat(Button): Extend icon moving to support Icon and react-icon children

### LINKED ISSUE #716: Button - Move icon children to the icon prop
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/issues/10663 and to help with https://github.com/patternfly/patternfly-react/issues/10698

In the empty state codemod I was able to identify icon children and move those to the icon prop, we should do that for buttons as well rather than only moving icons for plain buttons.

_Required actions:_
1. Build codemod
2. Build test

wise-king-sullyman:
Closes #716

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/buttonMoveIconsIconProp/button-moveIcons-icon-prop.test.ts
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/buttonMoveIconsIconProp/button-moveIcons-icon-prop.test.ts
@@ -81,5 +81,27 @@ ruleTester.run("button-moveIcons-icon-prop", rule, {
         },
       ],
     },
+    // with Icon component child
+    {
+      code: `import { Button, Icon } from '@patternfly/react-core'; <Button><Icon>Some icon</Icon></Button>`,
+      output: `import { Button, Icon } from '@patternfly/react-core'; <Button icon={<Icon>Some icon</Icon>}></Button>`,
+      errors: [
+        {
+          message: `Icons must now be passed to the \`icon\` prop of Button instead of as children. If you are passing anything other than an icon as children, ignore this rule when running fixes.`,
+          type: "JSXElement",
+        },
+      ],
+    },
+    // with react-icons icon child
+    {
+      code: `import { Button } from '@patternfly/react-core'; import { SomeIcon } from "@patternfly/react-icons"; <Button><SomeIcon /></Button>`,
+      output: `import { Button } from '@patternfly/react-core'; import { SomeIcon } from "@patternfly/react-icons"; <Button icon={<SomeIcon />}></Button>`,
+      errors: [
+        {
+          message: `Icons must now be passed to the \`icon\` prop of Button instead of as children. If you are passing anything other than an icon as children, ignore this rule when running fixes.`,
+          type: "JSXElement",
+        },
+      ],
+    },
   ],
 });
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/buttonMoveIconsIconProp/button-moveIcons-icon-prop.ts
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/buttonMoveIconsIconProp/button-moveIcons-icon-prop.ts
@@ -6,12 +6,15 @@ import {
   getAttributeValue,
   getExpression,
   getChildrenAsAttributeValueText,
+  getChildElementByName,
+  isReactIcon,
 } from "../../helpers";
 
 // https://github.com/patternfly/patternfly-react/pull/10663
 module.exports = {
   meta: { fixable: "code" },
   create: function (context: Rule.RuleContext) {
+    const source = context.getSourceCode();
     const { imports } = getFromPackage(context, "@patternfly/react-core");
 
     const buttonImport = imports.find(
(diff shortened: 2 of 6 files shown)


thatblindgeye (approved):

packages/eslint-plugin-pf-codemods/src/rules/v6/buttonMoveIconsIconProp/button-moveIcons-icon-prop.test.ts
```diff
@@ -81,5 +81,27 @@ ruleTester.run("button-moveIcons-icon-prop", rule, {
...
+    {
+      code: `import { Button } from '@patternfly/react-core'; import { SomeIcon } from "@patternfly/react-icons"; <Button><SomeIcon /></Button>`,
+      output: `import { Button } from '@patternfly/react-core'; import { SomeIcon } from "@patternfly/react-icons"; <Button icon={<SomeIcon />}></Button>`,
+      errors: [
+        {
+          message: `Icons must now be passed to the \`icon\` prop of Button instead of as children. If you are passing anything other than an icon as children, ignore this rule when running fixes.`,
+          type: "JSXElement",
+        },
+      ],
+    },
```
adamviktora, line 105: Can we also add a test for the children attribute mentioned above?

```suggestion
    },
    {
      code: `import { Button } from '@patternfly/react-core'; <Button variant="plain" children={<span>Some icon</span>} />`,
      output: `import { Button } from '@patternfly/react-core'; <Button icon={<span>Some icon</span>} variant="plain"  />`,
      errors: [
        {
          message: `Icons must now be passed to the \`icon\` prop of Button instead of as children. If you are passing anything other than an icon as children, ignore this rule when running fixes.`,
          type: "JSXElement",
        },
      ],
    },
```

packages/eslint-plugin-pf-codemods/src/rules/v6/buttonMoveIconsIconProp/button-moveIcons-icon-prop.ts
```diff
@@ -31,41 +34,60 @@ module.exports = {
...
                   node.children
                 );
               }
-              if (!childrenValue) {
+
+              if (!childrenValue && isPlain) {
+                return;
+              }
+
+              const iconComponentChild = getChildElementByName(node, "Icon");
```
adamviktora, line 74: I think we should check an `import { Icon }` from react-core, otherwise this will get any element named "Icon".

packages/eslint-plugin-pf-codemods/src/rules/v6/buttonMoveIconsIconProp/buttonMoveIconsIconPropInput.tsx
```diff
@@ -1,7 +1,18 @@
 import { Button } from "@patternfly/react-core";
```
adamviktora, line 1: Related to the comment above, an `Icon` import should be included here in order for the second element to have the rule applied.

packages/eslint-plugin-pf-codemods/src/rules/v6/buttonMoveIconsIconProp/button-moveIcons-icon-prop.ts
```diff
@@ -31,41 +34,60 @@ module.exports = {
...
+
+              const jsxElementChildren = node.children.filter(
+                (child) => child.type === "JSXElement"
+              ) as JSXElement[];
+              const reactIconChild = jsxElementChildren.find((child) =>
+                isReactIcon(context, child)
+              );
+
+              const iconChild = iconComponentChild || reactIconChild;
```
adamviktora, line 83: I wonder if we should also handle this icon child in case of children as an attribute. Because currently this is valid: 
```
import { Button } from '@patternfly/react-core'; 
import { SomeIcon } from "@patternfly/react-icons";
<Button children={<SomeIcon />} />
```
and the rule won't be applied.

I think it is a pretty rare use case and it would complicate the rule, so I wouldn't bother.

adamviktora (changes requested):
Couple of comments bellow, also there is a problem that this rule removes all children of the Button (because it was assuming only the plain button variant before), which becomes destructive in case like this:

```
import { Button } from "@patternfly/react-core";
import { SomeIcon } from "@patternfly/react-icons";

<Button>
  Text
  <SomeIcon />
</Button>
```
where the "Text" will be removed.

packages/eslint-plugin-pf-codemods/src/rules/v6/buttonMoveIconsIconProp/buttonMoveIconsIconPropInput.tsx
adamviktora, line 1:
```suggestion
import { Button, Icon } from "@patternfly/react-core";
```

packages/eslint-plugin-pf-codemods/src/rules/v6/buttonMoveIconsIconProp/buttonMoveIconsIconPropOutput.tsx
```diff
@@ -1,5 +1,12 @@
 import { Button } from "@patternfly/react-core";
```
adamviktora, line 1:
```suggestion
import { Button, Icon } from "@patternfly/react-core";
```

wise-king-sullyman: Ah yep, great catches!

adamviktora (approved):
🎉

PR #737, opened 2024-08-15
feat(UnavailableContent): update props to use bodyText

### LINKED ISSUE #732: [component-groups] - UnavailableContent
Follow up to breaking change https://github.com/patternfly/react-component-groups/pull/244

We've squashed the `preLinkBodyText` and `postLinkBodyText` into one `description` prop
 
Required actions:

Display a warning, that the old props have been removed and replaced with the new one

adamviktora:
Closes #732 

- because `unavailableBodyPreStatusLinkText` and `unavailableBodyPostStatusLinkText` has already been renamed to `preLinkBodyText` and `postLinkBodyText` and are being renamed again and merged to only one `bodyText` prop, I also modified the `component-groups-unavailable-content-rename-props` rule to only rename `unavailableTitleText`
- `statusPageLinkText` prop is now being used for a **primary** button instead of a link button -> the text should be capitalized (as it is being done with the default "status page" -> "Status page")


wise-king-sullyman (approved):
🔥

thatblindgeye (approved):

PR #738, opened 2024-08-16
fix(Card): updated markup for actionable cards

### LINKED ISSUE #736: Card - update markup/API has changed
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/10859

Based on changes made locally as followup to [card changes in Core](https://github.com/patternfly/patternfly/pull/6949), the markup for clickable cards is expected to change. Additionally, there will now be a console error if a clickable only or selectable only card does not have a proper accessible name (technically a change as it makes 1 of the aria props required, but this should have been how it was in the first place)

_Required actions:_
1. Build codemod
2. Build test

thatblindgeye:
Closes #736 

A todo is to update the message logic so that it'll only spit out the prop reference that is actually present (vs now it just spits out both `name` and `selectableActionId` props even if only one is present).

Also attempted a helper that could possibly be used in relation to https://github.com/patternfly/pf-codemods/issues/695


packages/eslint-plugin-pf-codemods/src/rules/helpers/nodeIsComponentNamed.ts
```diff
@@ -0,0 +1,9 @@
+import { JSXElement } from "estree-jsx";
+
+export function nodeIsComponentNamed(node: JSXElement, componentName: string) {
```
thatblindgeye, line 3: @adamviktora seems like you and I built similar helpers (this from you, `checkMatchingJSXOpeningElement` in the nodeMatches file from me)? Wdyt about us combining them into a single rule + if so, any preference on naming?

adamviktora: We can keep yours `checkMatchingJSXOpeningElement`, it is more general and we already have it as part of the initial general rule. I can refactor the two rules that are using this `nodeIsComponentNamed` to use `checkMatchingJSXOpeningElement` and then delete this helper.

adamviktora (commented):
Looking great! I really like the new helpers 🔥 

Some comments bellow regarding dealing with ImportDefaultSpecifier and few more

packages/eslint-plugin-pf-codemods/src/rules/helpers/getNodeForAttributeFixer.ts
```diff
@@ -0,0 +1,41 @@
...
+    return attribute.value;
+  }
+  if (
+    attribute.value.type === "JSXExpressionContainer" &&
+    ["ObjectExpression", "MemberExpression"].includes(
+      attribute.value.expression.type
+    )
+  ) {
+    return attribute.value.expression;
+  }
```
wise-king-sullyman, line 40: Nit, but since two of these three outcomes require the value type to be JSXExpressionContainer it could be refactored to simplify the logic a bit. Additionally refactoring to use a switch and deconstructing the expression out would also tidy things up a but.

Not a blocker though.

packages/eslint-plugin-pf-codemods/src/rules/v6/cardUpdatedClickableMarkup/card-updated-clickable-markup.test.ts
```diff
@@ -0,0 +1,138 @@
...
+    },
+    // 1 valid property + 2 invalid properties
+    {
+      code: `import { Card, CardHeader } from '@patternfly/react-core'; <Card isClickable><CardHeader selectableActions={{to: "#", name: 'Test', selectableActionId: 'Id'}} /></Card>`,
+      output: `import { Card, CardHeader } from '@patternfly/react-core'; <Card isClickable><CardHeader selectableActions={{to: "#"}} /></Card>`,
+      errors: [
+        {
+          message:
+            "The markup for clickable-only cards has been updated.Additionally, the `selectableActions.selectableActionId` and `selectableActions.name` props are no longer necessary to pass to CardHeader for clickable-only cards.",
+          // message: `The markup for clickable-only cards has been updated, now using button and anchor elements for the respective clickable action. The \`selectableActions.selectableActionId\` and \`selectableActions.name\` props are also no longer necessary for clickable-only cards. Passing them in will not cause any errors, but running the fix for this rule will remove them.`,
```
wise-king-sullyman, line 66: Did you intend to leave this in?

wise-king-sullyman (approved):

packages/eslint-plugin-pf-codemods/src/rules/v6/cardUpdatedClickableMarkup/card-updated-clickable-markup.test.ts
```diff
@@ -0,0 +1,138 @@
...
+          type: "JSXElement",
+        },
+      ],
+    },
+    // Passed as a variable reference
+    {
+      code: `import { Card, CardHeader } from '@patternfly/react-core'; const obj = {name: 'Test', selectableActionId: 'Id'}; <Card isClickable><CardHeader selectableActions={obj} /></Card>`,
```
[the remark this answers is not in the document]
thatblindgeye: That's actually a good point. In this particular case the variable may not be shared with anything other than a CardHeader component since it's such a specific object (though wouldn't be surprised). But for something like a prop that accepts a color, maybe that's a valid case where a variable is being used in other places (maybe something like a "color" variable that possibly shouldn't be updated if it's being used in other places).

If anything we can base it on a case-by-case basis, but worth keeping in mind when trying to apply a fix to a variable like this.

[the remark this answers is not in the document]
thatblindgeye: cc @wise-king-sullyman in case you had any thoughts

packages/eslint-plugin-pf-codemods/src/rules/v6/cardUpdatedClickableMarkup/card-updated-clickable-markup.test.ts
thatblindgeye: Whoops

adamviktora (approved):
LGTM 🎉

PR #739, opened 2024-08-19
feat(tokens): warn about old react-tokens usage

### LINKED ISSUE #733: Provide warning for consumers that are using our old CSS tokens
This pertains tot he topic of our new token documentation. We should at the very least have a codemod that will throw a warning if a consumer is using our old CSS tokens and to link them to any applicable token documentation.

It was noted during planning topics discussion that it can be dangerous for us to provide suggestions of what exact token a consumer should use instead, so it could be safer to at least get consumers to think about *why* they're using a specific CSS token and let them read through our token docs.

adamviktora:
Closes #733


packages/eslint-plugin-pf-codemods/src/rules/v6/tokensWarn/tokens-warn.ts
```diff
@@ -0,0 +1,3706 @@
...
+import { ImportDeclaration, ImportSpecifier } from "estree-jsx";
+
+module.exports = {
+  meta: {},
+  create: function (context: Rule.RuleContext) {
+    const tokensPackage = "@patternfly/react-tokens";
+
+    const { imports: tokenSpecifiers } = getFromPackage(context, tokensPackage);
+
+    const defaultTokensWithDeclaration = getDefaultImportsFromPackage(
```
wise-king-sullyman, line 16: Think this might be something worth helperizing?

wise-king-sullyman (approved):
Looks great! One non-blocking thought:

packages/eslint-plugin-pf-codemods/src/rules/v6/tokensWarn/tokens-warn.ts
adamviktora: I don't know if we ever use it again, I feel like it is quite specific. Do you have some potential future use case on your mind?

packages/eslint-plugin-pf-codemods/src/rules/v6/tokensWarn/tokens-warn.md
```diff
@@ -0,0 +1,17 @@
+### tokens-warn
+
+We have updated our CSS tokens. About half of our tokens have been replaced with newer ones. To find a suitable replacement token, check our new documentation https://staging-v6.patternfly.org/tokens/all-patternfly-tokens.
```
thatblindgeye, line 3: Let's wrap the link in some link text for this MD file; we can keep the url as-is in the message sent to the terminal

```suggestion
We have updated our CSS tokens. About half of our tokens have been replaced with newer ones. To find a suitable replacement token, check our [v6 token documentation](https://staging-v6.patternfly.org/tokens/all-patternfly-tokens).
```

thatblindgeye (commented):
Quick comment below, otherwise lgtm

packages/eslint-plugin-pf-codemods/src/rules/v6/tokensWarn/tokens-warn.ts
wise-king-sullyman: To clarify I don't mean for tokens explicitly, but something that returns the default imports file name and the import declaration in a nice easy to use format like this.

I feel like I've had a few places where it would've been nice so far, though I haven't dug through to find what those instances were. If you don't think it would likely be helpful it's always something we can revisit in the future if it turns out the other way.

thatblindgeye (approved):

packages/eslint-plugin-pf-codemods/src/rules/v6/tokensWarn/tokens-warn.ts
adamviktora: Yeah, you are right, I also ran to something similar maybe on the component-groups. I lean towards creating the helper as a followup, if we run to a problem where it would be helpful

adamviktora:
Also, we should maybe not only check react-tokens, but also raw strings in the code. 

We could also check CSS files, but that probably is not achievable by ESLint.

PR #740, opened 2024-08-19
feat(Tile): add deprecated import update mod

kmcfaul:
Codemod for https://github.com/patternfly/patternfly-react/pull/10821

Closes https://github.com/patternfly/pf-codemods/issues/731.

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/tileDeprecated/tile-deprecated.md
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/tileDeprecated/tile-deprecated.md
@@ -1,6 +1,6 @@
-### tile-deprecated [(#)](https://github.com/patternfly/patternfly-react/pull/10821)
+### tile-deprecated [(#10821)](https://github.com/patternfly/patternfly-react/pull/10821)
 
-Tile has been deprecated. Running the fix flag will update your imports to our deprecated package, but we suggest using Card instead.
+Tile has been deprecated. Running the fix flag will update your imports to our deprecated package, but we suggest using Card instead. There is a new Card example on our documentation showcasing how to set up a Card as a Tile.
 
 #### Examples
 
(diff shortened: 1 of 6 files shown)


packages/eslint-plugin-pf-codemods/src/rules/v6/tileDeprecated/tile-deprecated.md
```diff
@@ -0,0 +1,17 @@
+### tile-deprecated [(#)](https://github.com/patternfly/patternfly-react/pull/10821)
```
thatblindgeye, line 1: Just needs the PR number in the link text

```suggestion
### tile-deprecated [(#10821)](https://github.com/patternfly/patternfly-react/pull/10821)
```

packages/eslint-plugin-pf-codemods/src/rules/v6/tileDeprecated/tile-deprecated.md
```diff
@@ -0,0 +1,17 @@
...
+
+Tile has been deprecated. Running the fix flag will update your imports to our deprecated package, but we suggest using Card instead.
```
thatblindgeye, line 3: More of a nit since I don't think we've typically linked in this way, but wonder if it'd be worth linking to the Card as tile example once it goes in (via the v6 staging site)

packages/eslint-plugin-pf-codemods/src/rules/v6/tileDeprecated/tile-deprecated.test.ts
```diff
@@ -0,0 +1,22 @@
+const ruleTester = require('../../ruletester');
+import * as rule from './tile-deprecated';
+
+ruleTester.run('tile-deprecated', rule, {
+  valid: [
+    {
+      code: `import { Tile } from '@someOtherPackage';`,
+    },
+  ],
+  invalid: [
```
thatblindgeye, line 10: Can we just add the following tests:

- Tile import aliased
- Tile imports from dist (esm, js, and dynamic paths; the bannerReplaceVariantProp rule should have tests for reference)

thatblindgeye (changes requested):

thatblindgeye (approved):

adamviktora (approved):
LGTM

PR #745, opened 2024-08-21
feat: data-codemods cleanup rule

### LINKED ISSUE #723: Add codemod for removing our data-codemod attributes
Consumers would probably appreciate us not littering their code with our data tags, so we should give them a mod to remove them.

It should probably only run when manually called for using the `--only` flag, or something to that effect.

adamviktora:
Closes #723 

Also updates `baseReadme.md` file.

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/ruleCustomization.ts
+++ b/packages/eslint-plugin-pf-codemods/src/ruleCustomization.ts
@@ -2,7 +2,10 @@
  * add the rule name to the below array. Do not add rules here if the React PR is
  * not yet merged; instead add a "do not merge" label to the codemod PR.
  */
-export const betaRuleNames: string[] = ["kebabToggle-replace-with-menuToggle"];
+export const betaRuleNames: string[] = [
+  "data-codemods-cleanup",
+  "kebabToggle-replace-with-menuToggle",
+];
 
 // if you want a rule to have a severity that defaults to warning rather than error, add the rule name to the below array
 export const warningRules = [
--- a/packages/eslint-plugin-pf-codemods/src/ruleCustomization.ts
+++ b/packages/eslint-plugin-pf-codemods/src/ruleCustomization.ts
@@ -69,7 +69,6 @@ export const setupRules = [
 
 // rules that will run after other rules (cleanup imports?)
 export const cleanupRules = [
-  "data-codemods-cleanup",
   "no-unused-imports-v5",
   "no-unused-imports-v6",
   "no-duplicate-import-specifiers",
--- a/packages/eslint-plugin-pf-codemods/src/ruleCustomization.ts
+++ b/packages/eslint-plugin-pf-codemods/src/ruleCustomization.ts
@@ -69,6 +69,7 @@ export const setupRules = [
 
 // rules that will run after other rules (cleanup imports?)
 export const cleanupRules = [
+  "data-codemods-cleanup",
   "no-unused-imports-v5",
   "no-unused-imports-v6",
   "no-duplicate-import-specifiers",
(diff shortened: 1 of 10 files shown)


packages/eslint-plugin-pf-codemods/src/ruleCustomization.ts
```diff
@@ -69,6 +69,7 @@ export const setupRules = [
 
 // rules that will run after other rules (cleanup imports?)
 export const cleanupRules = [
+  "data-codemods-cleanup",
```
thatblindgeye, line 72: This will run the rule after every pass of a consumer running codemod fixes correct?

Wondering if we would want that to be the case. Using the v5 codemods that updated TableComposable to just Table and the then-currentl implementation of Table to TableDeprecated, if we removed a data-codemod attr from the TableComposable fixer, I imagine it would cause a scenario of TableComposable > updated to Table > updated to TableDeprecated incorrectly.

If that's the case, I feel like this rule might be better off as one that a consumer has to run explicitly, with us stating that they should only run it when they are finished running our codemods.

adamviktora: Yeah, you are right. I just tested that and on the first run of fixes it does everything correctly, but when you run the general codemod fixes again, it will show unwanted errors and do the unwanted fixes, due to the missing `data-codemods`. Which could lead to a lot of trouble, so yes, I'll change that rule to be runnable with only the `--only` option. Also it is working as v6 only rule right now, so that must change too.

packages/pf-codemods/index.js
```diff
@@ -81,6 +81,9 @@ function getRulesToRemove(options) {
 async function runCodemods(path, otherPaths, options) {
   const prefix = "@patternfly/pf-codemods/";
 
+  // data-codemods-cleanup rule won't be included in the recommended rule set
+  delete configs.recommended.rules[prefix + "data-codemods-cleanup"];
+
```
wise-king-sullyman, line 86: Could you not just add the rule to the betaRules array in the customization file instead of adding logic to this file?

packages/eslint-plugin-pf-codemods/src/rules/v6/dataCodemodsCleanup/dataCodemodsCleanupOutput.tsx
```diff
@@ -0,0 +1,13 @@
...
+  MastheadLogo,
+} from "@patternfly/react-core";
+
+export const DataCodemodsCleanupInput = () => (
+  <>
+    <DualListSelector />
+    <LoginMainFooterLinksItem />
+    <MastheadLogo />
+  </>
+);
```
wise-king-sullyman, line 13: Looks like the output file had the formatter ran on it after execution. Doesn't really matter now, but will if we ever implement auto checking of these tsx files in CI.

wise-king-sullyman (commented):
Looks great! Two small notes/thoughts:

packages/eslint-plugin-pf-codemods/src/rules/v6/dataCodemodsCleanup/dataCodemodsCleanupOutput.tsx
adamviktora: Oops, I thought we should run prettier on these output files, I do it every time 🤦

packages/pf-codemods/index.js
adamviktora: Oh, yeah, I did not read the betaRuleNames description, the name "beta" confused me a bit

thatblindgeye: FWIW Austin and I had a convo about the naming of the array, though I don't think we settled on whether to change it (or to what if so). Maybe post v6 we can have a brainstorm session about that + anything else we'd like to see changed in the repo

PR #747, opened 2024-08-21
feat(mod): add user-feedback css update warning

kmcfaul:
Closes https://github.com/patternfly/pf-codemods/issues/734.

I'm not sure if our codemods are meant to also cover extensions or not, but I figured it would be worth having a warning for `react-user-feedback` which was updated to need a separate `css` import.


packages/eslint-plugin-pf-codemods/src/rules/v6/userFeedbackWarnChanges/user-feedback-warn-changes.test.ts
```diff
@@ -0,0 +1,36 @@
...
+    {
+      code: `import { FeedbackModal } from '@patternfly/react-user-feedback';`,
+      output: `import { FeedbackModal } from '@patternfly/react-user-feedback';`,
+      errors: [
+        {
+          message: `FeedbackModal no longer internally references a scss stylesheet. You may have to import "Feedback.css" located in the dist "@patternfly/react-user-feedback/dist/esm/Feedback/Feedback.css" to maintain styling on FeedbackModal.`,
+          type: "ImportDeclaration",
+        },
+      ],
+    },
```
wise-king-sullyman, line 34: I'm a bit confused, what exactly is this second test testing that the first one isn't? They look the same?

packages/eslint-plugin-pf-codemods/src/rules/v6/userFeedbackWarnChanges/user-feedback-warn-changes.ts
```diff
@@ -0,0 +1,36 @@
+import { getFromPackage } from "../../helpers";
+import { Rule } from "eslint";
+import { ImportDeclaration } from "estree-jsx";
+
+// https://github.com/patternfly/react-user-feedback/pull/76
+module.exports = {
+  meta: {},
+  create: function (context: Rule.RuleContext) {
+    const { imports } = getFromPackage(context, "@patternfly/react-user-feedback");
```
wise-king-sullyman, line 9: It looks like user feedback supports default exports, so I think we probably need to check for those here too.

For this you should be able to just update your rule here to use the `getAllImportsFromPackage` helper instead of the `getFromPackage` helper, and then the `getDefaultDeclarationString` helper to handle those default imports. If you have any questions on that just lmk.

Probably also want to add a test for that.

wise-king-sullyman: Additionally I don't think this would catch any imports from `react-user-feedback/dist`, for that I think the quickest path would be to update `pfPackageMatches` to look for anything in user feedback like it does component groups.

And we should a test for that too.

wise-king-sullyman (changes requested):
Looks good! Just a few other import styles that I think we'd need to support and test for.

packages/eslint-plugin-pf-codemods/src/rules/v6/userFeedbackWarnChanges/user-feedback-warn-changes.test.ts
kmcfaul: When I checked another codemod I saw a similar setup, but I'll remove it. I'm not sure how it's testing the warning is only flagged a single time, looking at it again.

packages/eslint-plugin-pf-codemods/src/rules/v6/userFeedbackWarnChanges/user-feedback-warn-changes.ts
kmcfaul: I'm not sure what you mean by updating `pfPackageMatches`, that seems to be used in place of `getAllImportsFromPackage` which I'm currently using. Should I swap/adjust the import? 

I've also tried to adjust to account for the default import but my added (commented out) test isn't working - it's not returning anything from the `getAllImportsFromPackage` helper in the rule so no errors are reported.

@wise-king-sullyman

wise-king-sullyman (approved):

Issue #749, opened 2024-08-26
Issues encountered during early beta testing


nicolethoen:
[per this slack thread](https://redhat-internal.slack.com/archives/C04JTH3LQF6/p1724272824821169)

- [x] [TextListVariants and TextListItemVariants aren't updated via the existing codemod](https://github.com/patternfly/pf-codemods/pull/750) (Austin has PR up)
- [x] [ToolbarItem variants were removed and we need to remove them via codemod](https://github.com/patternfly/pf-codemods/pull/757)
- [x] [ToolbarItem had the `width` prop removed](https://github.com/patternfly/pf-codemods/pull/758)
- [x] [ToolbarGroup's variant prop and the ToolbarGroupVariant enum had the `button-group` value renamed](https://github.com/patternfly/pf-codemods/pull/759)

We should check that the codemods all work as expected or at least output some guidance for these situations.

Having some trouble with a few react-core components; TextListVariants, TextListItemVariants, EmptyStateHeader, and EmptyStateIcon.
![Screenshot 2024-08-26 at 3 28 58 PM](https://github.com/user-attachments/assets/09412901-d83b-4f3d-9c2f-4d1e53ba0be4)
![Screenshot 2024-08-26 at 3 24 31 PM](https://github.com/user-attachments/assets/ba90a488-e958-480b-a93d-62d2690cbe10)

ToolbarItem doesn't like serach-filter as a variant. And ToolbarContent complains about bulk-select 
![Screenshot 2024-08-26 at 3 30 18 PM](https://github.com/user-attachments/assets/91279afa-e52f-40e6-b620-227326f66c73)
![Screenshot 2024-08-26 at 3 32 11 PM](https://github.com/user-attachments/assets/a4a511e1-9948-4cbc-a842-99b7e57d45e9)

thatblindgeye:
FYI for the Toolbar updates, made in this PR in React: https://github.com/patternfly/patternfly-react/pull/10042

wise-king-sullyman:
The EmptyState changes do at least throw errors. It can't autofix because the EmptyStateHeader in use there doesn't have titleText or children.

thatblindgeye:
Leaving this open for now in case we decide to try providing an autofix for EmptyState when there's no titleText or children to use in the current autofix. Otherwise all other tasks are complete.

thatblindgeye:
Opened a separate issue for EmptyState, closing this issue as all other tasks are completed

PR #750, opened 2024-08-26
fix(Text): Replace missed text variant enums with content variant enum

wise-king-sullyman:
Not sure if it will close it, but at least towards #749


thatblindgeye (approved):
LGTM. Small nit so not a blocker at all, but teeny nice to have would be tests for the enums where we're passing them to a prop (rather than just a variable e.g. `<Content component={ContentVariant.ul} />`)

Issue #748, opened 2024-08-26
New CSS tokens - warn and do a mapping for global tokens


adamviktora:
Follow up to https://github.com/patternfly/pf-codemods/issues/733 

- [x] Provide warnings for Literals too (string like `"var(--pf-v5-whatever)"`) - DONE in https://github.com/patternfly/pf-codemods/pull/755
- [ ] Update every global token, which is not a color token - automatically via a fix.
- [ ] For colors, provide f.e. a hot pink color for every color so the consumers see it and then manually update to use the correct token.

PR #755, opened 2024-08-28
feat(tokens): warn for CSS variables in strings

adamviktora:
Towards #748 

Provides warnings also for variables in strings.

We might want to merge this now, so consumers have at least warnings, that this token is old and I'll work on the fixer as a followup.


packages/eslint-plugin-pf-codemods/src/rules/v6/tokensWarn/tokens-warn.ts
```diff
@@ -57,3650 +59,24 @@ module.exports = {
...
-  "global_secondary_color_100",
-  "global_spacer_form_element",
-  "global_success_color_100",
-  "global_success_color_200",
-  "global_target_size_MinHeight",
-  "global_target_size_MinWidth",
-  "global_warning_color_100",
-  "global_warning_color_200",
-];
+// consumers may run class-name-updater before codemods, so we have to check also old tokens with v6 prefix
```
thatblindgeye, line 77: Is this still applicable after changes made in https://github.com/patternfly/pf-codemods/pull/754 ?

thatblindgeye: > I would rather keep it there if someone has already run the old version of class-name-updater

I'm good with keeping it for this

PR #770, opened 2024-09-18
feat(EmptyState): make titleText optional

### LINKED ISSUE #764: EmptyState - enhance autofixer for when titleText and children are not present
Followup to https://github.com/patternfly/pf-codemods/issues/749. One of the issues found was that if there's no `titleText` or `children` for EmptyStateHeader, the autofixer we currently have doesn't run. We can try to provide an autofixer in this instance (the workaround consumer-side was passing `titleText=""` to EmptyState, which may not be ideal if we want to encourage consumers to pass a valid titleText).

adamviktora:
Closes #764

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/emptyStateHeaderMoveIntoEmptyState/emptyStateHeader-move-into-emptyState.ts
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/emptyStateHeaderMoveIntoEmptyState/emptyStateHeader-move-into-emptyState.ts
@@ -234,14 +234,16 @@ module.exports = {
             titleTextAttribute
           );
 
-          const titleText =
-            titleTextPropValue ||
-            (hasChildren
+          const createTitleTextPropFromChildren = () =>
+            hasChildren
               ? `titleText=${getChildrenAsAttributeValueText(
                   context,
                   header.children
                 )}`
-              : "");
+              : "";
+
+          const titleText =
+            titleTextPropValue || createTitleTextPropFromChildren();
 
           const iconPropValue = getExpression(headerIconAttribute?.value);
 
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/emptyStateHeaderMoveIntoEmptyState/emptyStateHeader-move-into-emptyState.ts
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/emptyStateHeaderMoveIntoEmptyState/emptyStateHeader-move-into-emptyState.ts
@@ -234,16 +234,14 @@ module.exports = {
             titleTextAttribute
           );
 
-          const createTitleTextPropFromChildren = () =>
-            hasChildren
-              ? `titleText=${getChildrenAsAttributeValueText(
-                  context,
-                  header.children
-                )}`
-              : "";
+          const childrenTitleText = hasChildren
+            ? `titleText=${getChildrenAsAttributeValueText(
+                context,
+                header.children
+              )}`
+            : "";
 
-          const titleText =
-            titleTextPropValue || createTitleTextPropFromChildren();
+          const titleText = titleTextPropValue || childrenTitleText;
 
           const iconPropValue = getExpression(headerIconAttribute?.value);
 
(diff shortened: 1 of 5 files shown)


thatblindgeye (approved):

packages/eslint-plugin-pf-codemods/src/rules/v6/emptyStateHeaderMoveIntoEmptyState/emptyStateHeader-move-into-emptyState.ts
```diff
@@ -269,10 +236,12 @@ module.exports = {
...
-            `titleText=${getChildrenAsAttributeValueText(
-              context,
-              header.children
-            )}`;
+            (hasChildren
+              ? `titleText=${getChildrenAsAttributeValueText(
+                  context,
+                  header.children
+                )}`
+              : "");
```
wise-king-sullyman, line 244: And one nit here: can you refactor this in a simpler way than having a ternary nested inside of an or statement?

packages/eslint-plugin-pf-codemods/src/rules/v6/emptyStateHeaderMoveIntoEmptyState/emptyStateHeader-move-into-emptyState.ts
```diff
@@ -204,6 +174,10 @@ module.exports = {
           "EmptyStateIcon"
         );
 
+        if (!header && !titleChild && !emptyStateIconChild) {
+          return;
+        }
+
```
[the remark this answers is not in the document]
adamviktora: This addition has to do with adding this valid test: `import { EmptyState } from "@patternfly/react-core"; <EmptyState>Foo bar</EmptyState>`, which was previously in the invalid array. I think we don't have to throw an error in such case (when consumers are using empty state without these props, so there is nothing to fix - well there would be a fix we don't want, adding an empty text, because `newEmptyStateProps` is empty), so that's why the early return. But it is a very rare case anyway.

packages/eslint-plugin-pf-codemods/src/rules/v6/emptyStateHeaderMoveIntoEmptyState/emptyStateHeader-move-into-emptyState.ts
```diff
@@ -231,19 +205,12 @@ module.exports = {
 
           hasTitleText = !!titleTextAttribute;
           hasIcon ||= !!headerIconAttribute;
-          hasChildren ||= header.children.length > 0;
+          hasChildren ||= !childrenIsEmpty(header.children);
```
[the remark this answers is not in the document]
adamviktora: Technically, if there are whitespaces, it is a JSXText children, so previously this: `<EmptyStateHeader> </EmptyStateHeader>` results to `hasChildren === true` and then we would add a titleText with a space as its value, which is not necessary. (Maybe prettier wouldn't allow it and it probably is a very rare situation)

packages/eslint-plugin-pf-codemods/src/rules/v6/emptyStateHeaderMoveIntoEmptyState/emptyStateHeader-move-into-emptyState.ts
```diff
@@ -234,14 +234,16 @@ module.exports = {
...
-            titleTextPropValue ||
-            (hasChildren
+          const createTitleTextPropFromChildren = () =>
+            hasChildren
               ? `titleText=${getChildrenAsAttributeValueText(
                   context,
                   header.children
                 )}`
-              : "");
+              : "";
```
wise-king-sullyman, line 243: One more even smaller nit, is there an advantage to this being a function rather than just doing something like

`const childrenTitleText = hasChildren ? <title text> : ""`

adamviktora: Oh, you are right, my thought around this was that if `titleTextPropValue` is not an empty string, then the second part of the OR operator won't be evaluated. But the function is super simple, so I guess it really doesn't make any difference performance-wise (and it is Javascript, so it's slow anyways)

wise-king-sullyman (approved):

PR #774, opened 2024-09-20
feat(tokens): prefix with t_

### LINKED ISSUE #771: Tokens - prefix with t_
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/11002

Prefix all react tokens, which are created from a token variable (in format `--pf-t--etc`), with a `t_`

_Required actions:_
1. Build codemod
2. Build test

adamviktora:
Closes #771


wise-king-sullyman (commented):
Looks great! Just need to update now that I merged #763

wise-king-sullyman (approved):
🥳

packages/eslint-plugin-pf-codemods/src/tokenLists/tTokens.ts
```diff
@@ -0,0 +1,856 @@
+// all React tokens in V6, whose value is a Patternfly token variable (has prefix --pf-t)
+// their naming was updated to be prefixed with 't_'
+export const tTokens = [
```
thatblindgeye, line 3: Super small nit, but wondering if this would make more sense named "reactTokens" or "tokensPrefixedWithT". Not a blocker and good to merge this as-is, justlet me know what you think.

adamviktora: yeah, tokensPrefixedWithT is good. I wouldn't use reactTokens, because also the pf-v6-whatever are react tokens

thatblindgeye (approved):

PR #779, opened 2024-09-30
feat(charts): import charts from @patternfly/react-charts/victory

dlabrecq:
This code mod updates existing import paths to our victory directory.

https://github.com/patternfly/patternfly-react/issues/11090

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/chartsImportMoved/chartsImport-moved.md
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/chartsImportMoved/chartsImport-moved.md
@@ -1,0 +1,19 @@
+### chartsImport-moved [(#11091)](https://github.com/patternfly/patternfly-react/pull/11091)
+
+In order to support multiple chart libraries, Victory based charts have moved. This rule will update import paths to our victory directory.
+
+Additionally, Victory has become a peer dependency.
+
+#### Examples
+
+In:
+
+```jsx
+%inputExample%
+```
+
+Out:
+
+```jsx
+%outputExample%
+```
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/chartsImportMoved/chartsImport-moved.test.ts
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/chartsImportMoved/chartsImport-moved.test.ts
@@ -1,0 +1,92 @@
+const ruleTester = require("../../ruletester");
+import * as rule from "./chartsImport-moved";
+import {
+  ValidTests,
+  InvalidTests,
+  createValidTest,
+  createInvalidTest,
+} from "../../helpers/testHelpers";
+
+const specifiersToMove = [
+  'Chart',
+  'ChartArea',
+  'ChartAxis',
+  'ChartBar',
+  'ChartBoxPlot',
+  'ChartBullet',
+  'ChartBulletComparativeErrorMeasure',
+  'ChartBulletComparativeMeasure',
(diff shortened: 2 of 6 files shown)


dlabaj (approved):
LGTM.  Thanks.

packages/eslint-plugin-pf-codemods/src/rules/v6/chartsImportMoved/chartsImport-moved.md
```diff
@@ -0,0 +1,19 @@
+### chartsVictory-promoted [(#11091)](https://github.com/patternfly/patternfly-react/pull/11091)
```
wise-king-sullyman, line 1: We should update the rule name here.

packages/eslint-plugin-pf-codemods/src/rules/v6/chartsImportMoved/chartsImport-moved.test.ts
```diff
@@ -0,0 +1,106 @@
...
+      `import {\n\t${specifier}\n} from '@patternfly/react-charts/victory';`,
+      [{ message: errorMessage, type: "ImportDeclaration" }]
+    )
+  );
+  invalidTests.push(
+    createInvalidTest(
+      `import { ${specifier} } from '@patternfly/react-charts/dist/dynamic/components/index.js';`,
+      `import {\n\t${specifier}\n} from '@patternfly/react-charts/victory';`,
+      [{ message: errorMessage, type: "ImportDeclaration" }]
+    )
```
wise-king-sullyman, line 99: Is it intended that for the cjs and dynamic imports it changes it to just `/victory`, but it keeps the esm import as explicitly esm?

packages/eslint-plugin-pf-codemods/src/rules/v6/chartsImportMoved/chartsImportMovedOutput.tsx
```diff
@@ -0,0 +1 @@
+import { Chart } from '@patternfly/react-charts/victory';
```
wise-king-sullyman, line 1: Nit: for the output files we like to have them as the 1:1 output of running the fixer, which here would have Chart on a new line.

The reasoning for this is that in the near future we will have a CI check that makes sure the output matches the input after autofixes are applied.

wise-king-sullyman (commented):
Thanks for putting this codemod together! Few small things:

packages/eslint-plugin-pf-codemods/src/rules/v6/chartsImportMoved/chartsImport-moved.md
dlabrecq: Yes, this should match. I renamed the files and didn't notice the comment

packages/eslint-plugin-pf-codemods/src/rules/v6/chartsImportMoved/chartsImportMovedOutput.tsx
dlabrecq: I don't really understand why Chart would be on a new line. I had to fix all that for Cost Management's linter.

packages/eslint-plugin-pf-codemods/src/rules/v6/chartsImportMoved/chartsImport-moved.test.ts
dlabrecq: I copied the Modal tests here. I'll update it

wise-king-sullyman (approved):

PR #780, opened 2024-10-01
chore(README): update README.md to include latest codemods

adamviktora:
Updates README.md

I had to include missing Input.tsx and Output.tsx in one of the rules in order for the `buildReadme.ts` script to work. Should I instead modify the script so the input / output example files are not required?


wise-king-sullyman (changes requested):
Yeah I think we should update the script so that it can handle rules without the input/output files. It isn't like they really do a lot of good for non-fix rules anyways.

wise-king-sullyman (approved):
🎉

PR #782, opened 2024-10-02
feat(MenuToggle): remove splitButtonOptions prop and interface

### LINKED ISSUE #781: [MenuToggle] - SplitButtonOptions variant prop
Follow up to breaking change issue https://github.com/patternfly/patternfly-react/issues/11095

SplitButtonOptions variant prop is no longer needed since the `pf-m-actions` modifier was removed

_Required actions:_
1. Build codemod
2. Build test

adamviktora:
Closes #781 

- also some helper updates


wise-king-sullyman (approved):
🚀

PR #797, opened 2024-10-23
fix(TextReplaceWithContent): update dist import path

### LINKED ISSUE #791: The Text -> Content codemod seems to not update dist import paths appropriately
Running the codemod updated the Text component import and usage appropriately, but it did not update the import path

Image

adamviktora:
Closes #791

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/textReplaceWithContent/text-replace-with-content.ts
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/textReplaceWithContent/text-replace-with-content.ts
@@ -71,11 +71,28 @@ module.exports = {
               const specifierName = specifierToReplace.imported.name;
               const newText = getNewText(specifierName);
 
+              const importPath = node.source.value?.toString() as string;
+
               context.report({
                 node,
                 message: errorMessage,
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin-pf-codemods/src/rules/v6/textReplaceWithContent/text-replace-with-content.ts
```diff
@@ -71,11 +71,28 @@ module.exports = {
               const specifierName = specifierToReplace.imported.name;
               const newText = getNewText(specifierName);
 
+              const importPath = node.source.value?.toString() as string;
```
wise-king-sullyman, line 74: Is there a way to do this without the type assertion? If not is this type assertion truly safe?

It seems like `importPath` could potentially resolve as undefined here.

adamviktora: I mean, technically it should always be a string starting with `"@patternfly/react-core"`, but I'll update it so there is not the casting

thatblindgeye (approved):

wise-king-sullyman (approved):
🚀

PR #799, opened 2024-11-04
feat(componentGroups): rename InvalidObjectProps to MissingPageProps

### LINKED ISSUE #786: Component groups InvalidObject rename does not update InvalidObjectProps
Our codemod for the InvalidObject component of component groups doesn't seem to be updating the interface export:

Image

adamviktora:
Closes #786 

Adds a helper to rename interface. I added the `@typescript-eslint/utils` package for that reason to type the given typescript Nodes. However it does it not fit 100% with the current `Rule.RuleContext` we are using, so I had to do the `node as unknown as Node` casting. 

In the future we might migrate to use `@typescript-eslint/utils` instead of `estree-jsx` entirely, but it would require basically updating every rule and helpers, so I don't know if it is worth it.

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/rules/helpers/renameInterface.ts
+++ b/packages/eslint-plugin-pf-codemods/src/rules/helpers/renameInterface.ts
@@ -1,10 +1,13 @@
 import { Rule } from "eslint";
 import { TSESTree } from "@typescript-eslint/utils";
-import { Identifier, ImportSpecifier } from "estree-jsx";
-import { getFromPackage } from "./getFromPackage";
-import { checkMatchingImportSpecifier } from "./nodeMatches/checkMatchingImportSpecifier";
+import { Identifier, ImportDeclaration, ImportSpecifier } from "estree-jsx";
+import {
+  checkMatchingImportSpecifier,
+  getFromPackage,
+  pfPackageMatches,
+} from ".";
 
-interface InterfaceRenames {
+interface Renames {
   [currentName: string]: string;
 }
 
@@ -13,11 +16,12 @@ function formatDefaultMessage(oldName: string, newName: string) {
 }
 
 export function renameInterface(
-  renames: InterfaceRenames,
+  interfaceRenames: Renames,
+  componentRenames: Renames,
   packageName = "@patternfly/react-core"
 ) {
   return function (context: Rule.RuleContext) {
-    const oldNames = Object.keys(renames);
+    const oldNames = Object.keys(interfaceRenames);
     const { imports } = getFromPackage(context, packageName, oldNames);
 
     if (imports.length === 0) {
@@ -38,7 +42,7 @@ export function renameInterface(
 
     const replaceIdentifier = (identifier: Identifier) => {
       const oldName = identifier.name;
-      const newName = renames[oldName];
+      const newName = interfaceRenames[oldName];
 
       context.report({
         node: identifier,
@@ -50,13 +54,33 @@ export function renameInterface(
     };
 
     return {
+      ImportDeclaration(node: ImportDeclaration) {
+        if (!pfPackageMatches(packageName, node.source.value)) {
+          return;
+        }
+        for (const oldName of Object.keys(componentRenames)) {
+          const newName = componentRenames[oldName];
+          const importSource = node.source.raw;
+          const importSourceHasComponentName = importSource?.includes(oldName);
+          const newImportDeclaration = importSource?.replace(oldName, newName);
+
+          if (newImportDeclaration && importSourceHasComponentName) {
+            context.report({
+              node,
+              message: formatDefaultMessage(oldName, newName),
+              fix: (fixer) =>
+                fixer.replaceText(node.source, newImportDeclaration),
+            });
+          }
+        }
+      },
       ImportSpecifier(node: ImportSpecifier) {
         if (!checkMatchingImportSpecifier(node, imports)) {
           return;
         }
 
         const oldName = node.imported.name;
-        const newName = renames[oldName];
+        const newName = interfaceRenames[oldName];
 
         context.report({
           node,
--- a/packages/eslint-plugin-pf-codemods/src/rules/helpers/renameInterface.ts
+++ b/packages/eslint-plugin-pf-codemods/src/rules/helpers/renameInterface.ts
@@ -1,0 +1,77 @@
+import { Rule } from "eslint";
+import { TSESTree } from "@typescript-eslint/utils";
+import { Identifier, ImportSpecifier, Node } from "estree-jsx";
+import { getFromPackage } from "./getFromPackage";
+
+interface InterfaceRenames {
+  [currentName: string]: string;
+}
+
+function formatDefaultMessage(oldName: string, newName: string) {
+  return `${oldName} has been renamed to ${newName}.`;
+}
+
+export function renameInterface(
+  renames: InterfaceRenames,
+  packageName = "@patternfly/react-core"
+) {
+  return function (context: Rule.RuleContext) {
+    const oldNames = Object.keys(renames);
+    const { imports } = getFromPackage(context, packageName, oldNames);
+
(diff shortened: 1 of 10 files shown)


packages/eslint-plugin-pf-codemods/src/rules/helpers/renameInterface.ts
```diff
@@ -0,0 +1,84 @@
...
+import { Identifier, ImportSpecifier, Node } from "estree-jsx";
+import { getFromPackage } from "./getFromPackage";
+
+interface InterfaceRenames {
+  [currentName: string]: string;
+}
+
+function formatDefaultMessage(oldName: string, newName: string) {
+  return `${oldName} has been renamed to ${newName}.`;
+}
```
thatblindgeye, line 12: Not a blocker, but wondering if we should add a directory to `helpers` for these sort of default message functions. `defaultMessages` as the directory -> `renameIdentifier.tsx` as this helper for example. Or just yanking this out as its own individual helper to be reused elsewhere.

packages/eslint-plugin-pf-codemods/src/rules/helpers/renameInterface.ts
```diff
@@ -0,0 +1,84 @@
...
+          return fixer.replaceText(identifier, newName);
+        },
+      });
+    };
+
+    return {
+      ImportSpecifier(node: ImportSpecifier) {
+        if (
+          !imports.some(
+            (specifier) => specifier.imported.name === node.imported.name
```
thatblindgeye, line 55: Could we either rework/rename the checkSpecifierExists function in the checkMatchingImportDeclaration file to be used here, or create a matcher helper specifically for a specifier only?

packages/eslint-plugin-pf-codemods/src/rules/helpers/renameInterface.ts
```diff
@@ -0,0 +1,84 @@
...
+          },
+        });
+      },
+      TSTypeReference(node: TSESTree.TSTypeReference) {
+        if (node.typeName.type === "Identifier") {
+          replaceIdentifier(node.typeName);
+        }
+      },
+      TSInterfaceHeritage(node: TSESTree.TSInterfaceHeritage) {
+        if (node.expression.type === "Identifier") {
```
thatblindgeye, line 78: Similar to above, maybe a matcher helper for these 2? I feel like rather than replaceIdentifier having a check to return early if an interface name isn't found in the imports, we should instead do what we're doing with the ImportSpecifier block and check first and only call the code to replace if a match is found.

packages/eslint-plugin-pf-codemods/src/rules/v6/componentGroupsInvalidObjectPropsRenameToMissingPageProps/component-groups-invalidObjectProps-rename-to-missingPageProps.md
```diff
@@ -0,0 +1,18 @@
+### component-groups-invalidObjectProps-rename-to-missingPageProps [(react-component-groups/#313)](https://github.com/patternfly/react-component-groups/pull/313)
```
thatblindgeye, line 1: This isn't a blocker for this PR, but sorta wondering if this should really be its own separate rule or included as part of the pre-existing one (with a rename maybe). Really the question is, "should there be a single rule dealing with anything InvalidObject rename related?"

Doing so would require a little refactoring both to the existing rule and the new helper here possibly, but also depends whether doing so would just bloat the existing rule too much. Having them separate may be more clear and understandable than trying to combine them, but curious what you think.

packages/eslint-plugin-pf-codemods/src/rules/v6/componentGroupsInvalidObjectPropsRenameToMissingPageProps/component-groups-invalidObjectProps-rename-to-missingPageProps.test.ts
```diff
@@ -0,0 +1,95 @@
...
+        code: `const props: InvalidObjectProps;`,
+      },
+      // import from wrong package
+      {
+        code: `import { InvalidObjectProps } from '@patternfly/react-core';`,
+      },
+    ],
+    invalid: [
+      {
+        code: `import { InvalidObjectProps } from '@patternfly/react-component-groups';
```
thatblindgeye, line 22: Maybe a valid test for importing something else from the correct package, and an invalid test that imports InvalidObjectProps and something else to show that only InvalidObjectProps gets touched/flagged?

thatblindgeye (changes requested):

packages/eslint-plugin-pf-codemods/src/rules/v6/componentGroupsInvalidObjectPropsRenameToMissingPageProps/component-groups-invalidObjectProps-rename-to-missingPageProps.md
adamviktora: > "should there be a single rule dealing with anything InvalidObject rename related?"
I think ideally there should be just a single rule, but...

> Doing so would require a little refactoring
I am afraid it could end up being "a lot of" refactoring, given that the current `renameComponent` helper already returns a function. But now that I am thinking about it, we could combine it and do something like:
```
create: (context: Rule.RuleContext) => {...renameComponent(context), ...renameInterface(context)}
```
But maybe it will not be that easy

packages/eslint-plugin-pf-codemods/src/rules/helpers/renameInterface.ts
adamviktora: I updated this to have a separate function to check whether the renaming should be done.

Do you see a reason in extracting that to a separate helper as well?

packages/eslint-plugin-pf-codemods/src/rules/v6/componentGroupsInvalidObjectPropsRenameToMissingPageProps/component-groups-invalidObjectProps-rename-to-missingPageProps.test.ts
```diff
@@ -0,0 +1,99 @@
...
+        errors: [
+          {
+            message,
+            type: "ImportSpecifier",
+          },
+        ],
+      },
+      {
+        code: `import { InvalidObjectProps } from '@patternfly/react-component-groups/dist/dynamic/InvalidObject';`,
+        output: `import { MissingPageProps } from '@patternfly/react-component-groups/dist/dynamic/InvalidObject';`,
```
wise-king-sullyman, line 89: Shouldn't these dist imports be updated to `dynamic/MissingPageProps`?

thatblindgeye (approved):

packages/eslint-plugin-pf-codemods/src/rules/v6/componentGroupsInvalidObjectPropsRenameToMissingPageProps/component-groups-invalidObjectProps-rename-to-missingPageProps.test.ts
adamviktora: Good catch!

I wanted to fix this in the `renameComponent` helper, so it updates the import path no matter which component is imported (so it fixes things like `import { someHelper } from "path/OldButton"` to `import { someHelper } from "path/NewButton"`, currently only JSXOpeningElements are handled (so only when `OldButton` is imported from `"path/OldButton"`). Such a change assumes that whenever a coponent is renamed, also its directory is renamed.

I had the change ready, but it broke too many tests (we would need to update them to also add error for "ImportDeclaration").

So for now I just added a workaround to the `renameInterface` helper. I don't love it but it will do the job.

wise-king-sullyman (approved):
🔥

PR #803, opened 2024-11-14
fix: class name updater does not support absolute paths

### LINKED ISSUE #802: v6 class name updater does not support absolute paths
Passing absolute paths to the class name updater causes an error to be thrown.

It appears that the absolute path is incorrectly being resolved to being relative to the dir where the command is ran.

Image

rh-gvincent:
Closes #802

If the path given to class name updater is an absolute path, the tool raises an error. It adds the current working dir regardless of the path type.

This patch fixes this issue and only adds the current working dir if the file path is not absolute.


wise-king-sullyman (approved):
🚀

PR #811, opened 2025-01-10
Helper documentation/restructure POC

thatblindgeye:
Towards https://github.com/patternfly/pf-codemods/issues/722 and https://github.com/patternfly/pf-codemods/issues/725.

For the documentation, we can build upon jsdocs basically. The main thing is the description of the helper function: it should include where it is typically called (inline, at the top of a specific block, as a reference to the create method, etc), and the typical use case ("paired with other logic to conditionally do X or Y").

For the restructuring, I went with the intent of separating things out into their own files as much as possible, then from there creating directories for things that belong together. In this draft for example, the testHelpers file is split into 3 separate files with each housed in a testHelpers directory (the JSXAttributes file would be split out similarly, I only added some documentation there for now). The `getImportPath` helper, though, may not really fit in with any other helpers and would remain in the main `helpers` directory.

One question could be what we want to do with things that may not be used elsewhere, e.g. the checkMatchingImportDeclaration helper has a `checkSpecifierExists` function that's only used for that helper. We could just move those things to their own file as well -- worse case it's an extra file, best case maybe that file itself can be used elsewhere as well. I don't have a strong opinion either way.


packages/eslint-plugin-pf-codemods/src/rules/helpers/JSXAttributes.ts
```diff
@@ -35,6 +47,25 @@ export function getAnyAttribute(
   return foundAttribute;
 }
 
+/**
+ * Use inline to pair with other logic (such as getting attribute values or calling logic based on an attribute being present or not) and when you need to get several attributes at once.
+ * @param node - The node in which to find the specified attributes.
+ * @param attributeNames - An array of attribute names to find on the node.
+ * @returns An object containing key:value pairs of the attribute names and either their found JSXAttribute object or undefined.
+ */
+export function getSpecifiedAttributes(
```
adamviktora, line 56: 👍 👍

packages/eslint-plugin-pf-codemods/src/rules/helpers/nodeMatches/checkMatchingImportDeclaration.ts
```diff
@@ -1,21 +1,34 @@
...
   );
 }
 
-/** Used to check whether the current ImportDeclaration node matches at least 1 of the import specifiers. */
+/**
+ * Use inline at the start of an ImportDeclaration block to early return if no specifiers are found, or only run logic if at least 1 specifier is found.
+ * @param node - The node to check for any import specifiers.
+ * @param imports - A single import specifier or an array of specifiers to find within the import declaration.
+ * @param packageName - The package name to check against the import declaration source.
+ * @returns A boolean depending on whether the import declaration source matches the packageName, or whether at least 1 import specifier is found.
```
adamviktora, line 24: Maybe I'd rewrite this to `A boolean depending on whether the import declaration source matches the packageName and at least 1 import specifier is found.` So it doesn't sound like it can return true if only the packageName matches. (but maybe it is just my Czech->English language barrier and it makes sense)

packages/eslint-plugin-pf-codemods/src/rules/helpers/testHelpers/createInvalidTest.tsx
```diff
@@ -0,0 +1,20 @@
+import { TestErrors } from "./testingTypes";
+
+/**
+ * Use inline in a test file to push a test object to an invalid test array.
+ * @param code The invalid code snippet that should trigger an error.
+ * @param output The code snippet that should be output upon the fixer being ran. If no fixer exists, this shoujld be the same as the code parameter.
```
adamviktora, line 6: nit: typo
```suggestion
 * @param output The code snippet that should be output upon the fixer being ran. If no fixer exists, this should be the same as the code parameter.
```

adamviktora (commented):
Great job on this!

I really like the new JSDocs and the description sentences structure.

Separating each function to its own file and then grouping similar helpers to folders seems like a good approach.

> We could just move those things to their own file as well -- worse case it's an extra file, best case maybe that file itself can be used elsewhere as well

I agree with that.

wise-king-sullyman:
> One question could be what we want to do with things that may not be used elsewhere

I agree with us still splitting them out.

On the jsdoc stuff I kind of wonder if the param and return descriptions are really adding much for value in most cases?

I think my reservations there kind of stem from the fact that TS is already showing _what_ a function takes/returns, and it can't become outdated in the way written jsdoc style descriptions can. I _think_ that I feel like the combo of the jsdoc general description sentence and the types from TS are enough. But I'm not dead set.

thatblindgeye:
> On the jsdoc stuff I kind of wonder if the param and return descriptions are really adding much for value in most cases?

That's something I was wondering as well, seemed like most of the time we may not be adding much value that wouldn't be evident from the TS info. We could make the descriptions optional unless there's info that needs to/should be known that TS won't immediately convey, or if we want to provide more up front information that may not be totally clear from the TS types (I think Adam has some potential examples of this in https://github.com/patternfly/pf-codemods/pull/796/files)

PR #822, opened 2025-02-07
feat(ExpandableSection): Remove isActive prop

### LINKED ISSUE #815: [Expandable Section] - [isActive prop was removed]
Follow up to breaking change PR https://github.com/patternfly/patternfly-react/pull/9881

The `isActive prop was removed for Penta.  It is no longer needed.

_Required actions:_
1. Build codemod
2. Build test

mfrances17:
Closes #815.

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/expandableSectionRemoveIsActiveProp/expandableSection-remove-isActive-prop.test.ts
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/expandableSectionRemoveIsActiveProp/expandableSection-remove-isActive-prop.test.ts
@@ -11,9 +11,43 @@ ruleTester.run("expandableSection-remove-isActive-prop", rule, {
     },
   ],
   invalid: [
+    // importing from react-core
     {
       code: `import { ExpandableSection } from '@patternfly/react-core'; <ExpandableSection isActive />`,
       output: `import { ExpandableSection } from '@patternfly/react-core'; <ExpandableSection  />`,
+      errors: [
+        {
+          message: `The \`isActive\` prop has been removed from ExpandableSection.`,
+          type: "JSXOpeningElement",
+        },
+      ],
+    },
+    // importing from dist/dynamic
+    {
+      code: `import { ExpandableSection } from '@patternfly/react-core/dist/dynamic/components/ExpandableSection'; <ExpandableSection isActive />`,
+      output: `import { ExpandableSection } from '@patternfly/react-core/dist/dynamic/components/ExpandableSection'; <ExpandableSection  />`,
+      errors: [
+        {
+          message: `The \`isActive\` prop has been removed from ExpandableSection.`,
+          type: "JSXOpeningElement",
+        },
+      ],
+    },
+    // importing from dist/esm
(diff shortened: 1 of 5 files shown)


packages/eslint-plugin-pf-codemods/src/rules/v6/expandableSectionRemoveIsActiveProp/expandableSection-remove-isActive-prop.test.ts
```diff
@@ -0,0 +1,25 @@
...
+ruleTester.run("expandableSection-remove-isActive-prop", rule, {
+  valid: [
+    {
+      code: `<ExpandableSection isActive />`,
+    },
+    {
+      code: `import { ExpandableSection } from '@patternfly/react-core'; <ExpandableSection someOtherProp />`,
+    },
+  ],
+  invalid: [
```
thatblindgeye, line 13: One thing we typically try to do is include tests where the import path is `@patternfly-react/dist/[esm|js|dynamic]/...` (the dualListSelectorNextPromoted rule has such tests). Would you mind just adding those in here as well? Everything else in the test can remain the same as this one, just the import path getting updated.

packages/eslint-plugin-pf-codemods/src/rules/v6/expandableSectionRemoveIsActiveProp/expandableSectionRemoveIsActivePropOutput.tsx
```diff
@@ -0,0 +1,5 @@
+import { ExpandableSection } from "@patternfly/react-core";
+
+export const ExpandableSectionRemoveIsActivePropInput = () => (
+  <ExpandableSection isActive />
```
thatblindgeye, line 4: This file should have the prop removed and match the expected output (additional white space if there is one).

thatblindgeye (changes requested):

packages/eslint-plugin-pf-codemods/src/rules/v6/expandableSectionRemoveIsActiveProp/expandableSectionRemoveIsActivePropOutput.tsx
mfrances17: done, ty

packages/eslint-plugin-pf-codemods/src/rules/v6/expandableSectionRemoveIsActiveProp/expandableSection-remove-isActive-prop.test.ts
mfrances17: added, when in development i didn't see many codemods that do that, at least not consistently, but now that i revisit i do see a few. easy enough to do, thanks.

thatblindgeye: Yeah to be fair it's something we intend to go back to to ensure they all do that, so it's more a "if we remember to we include them" rule 😆

PR #823, opened 2025-02-07
fix(chipGroup): Replace numChips with numLabels

### LINKED ISSUE #818: Chip Group codemod: numChips is not being changed to numLabels
After running the codemods, the `numChips` prop for LabelGroup is not  updated to `numLables`

mfrances17:
Closes #818

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/chipReplaceWithLabel/chip-replace-with-label.test.ts
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/chipReplaceWithLabel/chip-replace-with-label.test.ts
@@ -26,18 +26,18 @@ ruleTester.run('chip-replace-with-label', rule, {
       errors: [chipImportError],
     },
     {
-      code: `import { ChipGroup } from '@patternfly/react-core/deprecated'; <ChipGroup someOtherProp>This is a chipgroup</ChipGroup>`,
-      output: `import { LabelGroup } from '@patternfly/react-core'; <LabelGroup someOtherProp>This is a chipgroup</LabelGroup>`,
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin-pf-codemods/src/rules/v6/chipReplaceWithLabel/chip-replace-with-label.test.ts
```diff
@@ -26,30 +26,30 @@ ruleTester.run('chip-replace-with-label', rule, {
       errors: [chipImportError],
     },
     {
-      code: `import { ChipGroup } from '@patternfly/react-core/deprecated'; <ChipGroup someOtherProp>This is a chipgroup</ChipGroup>`,
-      output: `import { LabelGroup } from '@patternfly/react-core'; <LabelGroup someOtherProp>This is a chipgroup</LabelGroup>`,
+      code: `import { ChipGroup } from '@patternfly/react-core/deprecated'; <ChipGroup numChips={3} someOtherProp>This is a chipgroup</ChipGroup>`,
```
thatblindgeye, line 29: Rather than adding the old prop on several other existing tests, can we instead create a new test that tests specifically for the numChips prop? Similar to the test for isReadOnly prop

mfrances17: i think the new tests that i added are valid test cases... so instead i reverted the more isolated tests to their originals, added an isolated numChips test, and kept the new tests as additional tests

thatblindgeye (approved):

PR #834, opened 2025-03-12
fix: updated the cache to v4 due to EOL of v2.

dlabaj:
Update from v2 to v4.


.github/workflows/release.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       GH_TOKEN: ${{ secrets.GH_BRANCH_PROTECTION }} # needs to be an admin token to get around branch protection
       GH_PR_TOKEN: ${{ secrets.GH_PR_TOKEN }}
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
```
wise-king-sullyman, line 17: Are the updates to the checkout action here (and in the other file) intentional?

wise-king-sullyman (approved):

PR #848, opened 2025-07-17
fix: split table hasAnimations fix with cli option

gitdallas:
if a user passes the option to only run enable-animations, it will skip the version check and ask them "This will update several React Core components. Which components would you like to update? (Note: Some users have reported issues with enabling animations on Table.)"

i thought about adding "If an issue happens, it's Eric's fault", but didn't.

with choices: 
"Just React Core components"
or
"React Core and Table components"


gitdallas:
hang on while i fix a test 😬

wise-king-sullyman (approved):

PR #854, opened 2025-08-11
feat(853): favorite button codemod

### LINKED ISSUE #853: Button - support isFavorite and isFavorited on Button for animation styling
Noted by @andrew-ronaldson. The existing implementation of Button in RHOAI does not apply the new variant that supports animations. cc @gitdallas @thatblindgeye 

See https://www.patternfly.org/components/button/#favorite 

Since this variant is [newly supported](https://github.com/patternfly/patternfly-react/pull/11853) in PF, RHOAI was a using [custom implementation of a `Button`](https://github.com/opendatahub-io/odh-dashboard/blob/4b8c2bf15f0bbdce4a757f661aece5ef62eb81ca/frontend/src/components/FavoriteButton.tsx#L13-L27) that uses an icon prop to directly apply favorite icon + styling to the className. 

We need a codemod that can check for `StarIcon` and replace it with the new props for the favorite variant. We can also surface a warning for more complex cases.

gitdallas:
closes #853

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/ruleCustomization.ts
+++ b/packages/eslint-plugin-pf-codemods/src/ruleCustomization.ts
@@ -13,6 +13,7 @@ export const warningRules = [
   "aboutModalBoxHero-remove-subcomponent",
   "accordionItem-warn-update-markup",
   "applicationLauncher-warn-input",
+  "button-support-favorite-variant",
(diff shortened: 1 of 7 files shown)


jenny-s51 (approved):
Thanks @gitdallas , LGTM

PR #855, opened 2025-09-05
fix issue with valid log-viewer cssVars being removed

gitdallas:
came in up this chat: https://patternfly.slack.com/archives/C4FM977N0/p1756991427294309

likely they were removed based on patternfly.css differences, but these styles were just moved into another repo vs removed


mcoker (approved):
LGTM! Thanks @gitdallas!

PR #857, opened 2025-09-24
856: add note for AnimationsProvider in enable-animations docs

### LINKED ISSUE #856: AnimationsProvider - add documentation that AnimationsProvider is now supported
Follow up to PR: https://github.com/patternfly/patternfly-react/pull/11990

For visibility, add documentation that we now support a context provider by default, but leave the codemod the same for users who want to enable `hasAnimations` manually on components without context. 

We should also update the [README](https://github.com/patternfly/pf-codemods/blob/4519b808983f615fd178ff6b2bc70e22dc0b650c/packages/pf-codemods/README.md?plain=1#L1218-L1234) and https://github.com/patternfly/pf-codemods/blob/4519b808983f615fd178ff6b2bc70e22dc0b650c/packages/eslint-plugin-pf-codemods/src/rules/v6/enableAnimations/enable-animations.md to specify that the AnimationsProvider is available to use as an option as well.

gitdallas:
closes: https://github.com/patternfly/pf-codemods/issues/856

### DIFF
--- a/packages/eslint-plugin-pf-codemods/src/rules/v6/enableAnimations/enable-animations.md
+++ b/packages/eslint-plugin-pf-codemods/src/rules/v6/enableAnimations/enable-animations.md
@@ -2,6 +2,8 @@
 
 This rule adds the `hasAnimations` prop to PatternFly components that support animations. This is an optional enhancement that enables smoother transitions and animations in your UI components.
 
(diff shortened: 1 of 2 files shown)


jenny-s51 (approved):
Thank you @gitdallas 🚀

